PR #10, opened 2020-07-07
Create.stdhooks.new.rst

LP-Codes:
Add a hook for ``HtmlTestRunner``, which has a hidden import Template which fixes Error While Using With UnitTest Module.


Legorooj:
@LP-Codes please add this to your other PR, not a new one. You also need to call this file `8.new.rst`, not `stdhooks.new.rst`.

PR #17, opened 2020-07-30
add hook for workflow

zhouronghua:
need to hook for workflow

### DIFF
--- a/19.new.rst
+++ b/19.new.rst
@@ -1,0 +1,1 @@
+Add a hook for ``workflow``, which is missing version information contained in metadata.
--- a/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-workflow.py
+++ b/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-workflow.py
@@ -1,4 +1,3 @@
-#-----------------------------------------------------------------------------
 # Copyright (c) 2013-2020, PyInstaller Development Team.
 #
 # Distributed under the terms of the GNU General Public License (version 2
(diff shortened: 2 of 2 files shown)


bwoodsend:
Ahh I see this is already merged. @Legorooj Would it be easier to revert and ask zhouronghua to correct the name or just sort it ourselves?

Legorooj:
I didn't notice that, so good catch, but the number is actually semi-valid: #19. There's probably no point changing it now, cause that issue links back to this PR (albeit in a roundabout manner).

In this case, amending the commit locally and pushing is probably the easiest thing - I enabled force pushing to this repository. When you get write access to the main PyInstaller repository remember you _can't_ force push, so no history rewriting. Check 5 times more than necesarry before merging something on the main repo.

zhouronghua:
> Ahh I see this is already merged. @Legorooj Would it be easier to revert and ask zhouronghua to correct the name or just sort it ourselves?

sorry, not familier with the workflow, i thought the number was the issue i just generate, but in fact it is the pull request no

Legorooj:
@zhouronghua issues and PRs share the numbers, so if there's no issue for it already, then just use the PR number.

zhouronghua:
> @zhouronghua issues and PRs share the numbers, so if there's no issue for it already, then just use the PR number.
i get it, thank u

bwoodsend:
There we go. Sorted now. @zhouronghua For future reference - the `xxx.new.rst` file goes in the `news/` folder rather than the root. Other than that - good work!

PR #26, opened 2020-08-09
Add iminuit hooks

jeremysanders:
iminuit uses a cython module which imports sub-modules. These are not normally detected by pyinstaller. This hook uses collect_submodules to add the missing modules as hidden imports, excluding tests.


jeremysanders:
Thanks - is this ok?

bwoodsend:
Yeah, that's what Legorooj meant. I really should have spotted that the 1st time round.

PR #29, opened 2020-08-12
Add pyav hooks

liupx1988:
pyav uses a cython module which imports sub-modules. These are not
normally detected by pyinstaller. This hook uses collect_submodules
to add the missing modules as hidden imports.


liupx1988:
It seems test failed while tests/scripts is missing.

bwoodsend:
Yeah, the tests need some surgery. But your stuff seems to be passing. Nice work. I'll let @Legorooj have a final check before merging...

PR #32, opened 2020-08-16
Add missing licences to MANIFEST.in

### LINKED ISSUE #31: License missing in sdist uploaded to PyPI
There is no license file in https://pypi.io/packages/source/p/pyinstaller-hooks-contrib/pyinstaller-hooks-contrib-2020.6.tar.gz

bwoodsend:
Fixes #31 - i.e. we forgot to add the licenses to the `MANIFEST.in` so that they are included in `sdist`s. @Legorooj Can you confirm if both the `LICENSE` and `LICENSE.APL.txt` should be included or just the former? And does this warrant a changelog entry?

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,1 +1,3 @@
 graft src/_pyinstaller_hooks_contrib
+include LICENSE
+include LICENSE.APL.txt


Legorooj:
Huh. Good catch. And no it doesn't warrant an entry; the end users probably won't notice.

nehaljwani:
I am confused - why add only `LICENSE.APL.txt` and not `LICENSE.GPL.txt` ?

bwoodsend:
Ughh, at least @nehaljwani is awake. I've amended this on my branch but I see it's too late now the PR is merged. @Legorooj What's the cleanest way to deal with this?

Legorooj:
@bwoodsend amend the commit and force push. I should have noticed as well :rolls_eyes:

bwoodsend:
Hopefully... that's that sorted.

PR #34, opened 2020-08-18
Add hook for the pysnmp_mibs package

JoshDavisChem:
Hook for the pynsmp_mibs package at
https://pypi.org/project/pysnmp-mibs/

It is a package that contains a wide variety of MIB's precompiled for the pysnmp package

### DIFF
--- a/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-pysnmp_mibs.py
+++ b/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-pysnmp_mibs.py
@@ -1,0 +1,15 @@
+# ------------------------------------------------------------------
+# Copyright (c) 2020 PyInstaller Development Team.
+#
+# This file is distributed under the terms of the GNU General Public
+# License (version 2.0 or later).
+#
+# The full license is available in LICENSE.GPL.txt, distributed with
+# this software.
+#
+# SPDX-License-Identifier: GPL-2.0-or-later
+# ------------------------------------------------------------------
+
+from PyInstaller.utils.hooks import collect_data_files
+
+datas = collect_data_files('pysnmp_mibs', include_py_files=True)


bwoodsend:
@JoshDavisChem For the benefit of those of us who've never heard of this package or _MIBs_ before (like me) could you give a brief overview of what the issue with this package is (without this hook). It appears to be pure Python which usually means no issues. 

I'm mostly interested in how these `xxx-MIB.py` files are loaded? I take it that it is done by filename rather than Python's native `from pysnmp_mibs import xxx`? This would explain why you're treating python files as data files instead of hidden imports.

bwoodsend:
I should also say... thanks for working on this.

news/34.new.rst
```diff
@@ -0,0 +1 @@
+added hook for ``pysnmp_mibs`` for hidden data imports
```
bwoodsend, line 1:
```suggestion
Add hook for ``pysnmp_mibs`` which contains hidden data imports.
```
Always write in present tense. Same goes for commit messages.

src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-pysnmp_mibs.py
```diff
@@ -0,0 +1,18 @@
...
+# License (version 2.0 or later).
+#
+# The full license is available in LICENSE.GPL.txt, distributed with
+# this software.
+#
+# SPDX-License-Identifier: GPL-2.0-or-later
+# ------------------------------------------------------------------
+
+# pysnmp_mibs provides MIBS files for pysnmp. It contains MIB files
+# expanded beyond the default in pysnmp.
```
bwoodsend, line 14: Huh? _expanded beyond the default_? Could you rephrase? Perhaps:
```suggestion
# pysnmp_mibs provides MIBS files for pysnmp. It contains MIB files
# '.py' files which are treated as data files and are loaded by [what are they loaded by?]
```

```

JoshDavisChem:
This is still on my radar. I just have to go back to my original program and other documentation to review my comments and make them more specific.

Legorooj:
Any update on this?

PR #35, opened 2020-08-18
Create 34.new.rst

JoshDavisChem:


bwoodsend:
I take it this is actually meant for PR #34. In which case can you just add it to that one. Use:
```git
git checkout pysnmp_mibs
git checkout patch-1 -- news/34.new.rst
git add news/34.new.rst
git commit -m "Add news for PR 34"
git push https://github.com/JoshDavisChem/pyinstaller-hooks-contrib.git pysnmp_mibs
```

PR #40, opened 2020-08-28
Add a hook for publicsuffix2

mhils:
And here's the next one! 😃

I will send PRs sequentially, as there's otherwise a merge conflict on each PR.

### DIFF
--- a/src/_pyinstaller_hooks_contrib/tests/test_libraries.py
+++ b/src/_pyinstaller_hooks_contrib/tests/test_libraries.py
@@ -323,4 +323,5 @@ def test_passlib(pyi_builder):
 def test_publicsuffix2(pyi_builder):
     pyi_builder.test_source("""
         import publicsuffix2
+        publicsuffix2.PublicSuffixList()
         """)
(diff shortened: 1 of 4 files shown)


bwoodsend:
> I will send PRs sequentially, as there's otherwise a merge conflict on each PR.

Yes, please do...

src/_pyinstaller_hooks_contrib/tests/test_libraries.py
```diff
@@ -317,3 +317,10 @@ def test_passlib(pyi_builder):
     pyi_builder.test_source("""
         import passlib.apache
         """)
+
+
+@importorskip('publicsuffix2')
+def test_publicsuffix2(pyi_builder):
+    pyi_builder.test_source("""
+        import publicsuffix2
```
bwoodsend, line 325: This doesn't appear to be enough. I can compile and run this without the hook and it still works. Try:
```suggestion
        import publicsuffix2
        publicsuffix2.PublicSuffixList()
```
That gives me the error without the hook.

bwoodsend (changes requested):

src/_pyinstaller_hooks_contrib/tests/test_libraries.py
mhils: Good point, thanks. How do you go on to test this? I naively expected the pyi_builder thing to throw an error if execution succeeds in absence of a hook.

bwoodsend (approved):
Perfect

bwoodsend:
As usual, @Legorooj gets the final say on this...

Legorooj (approved):
Nice work and thanks!

PR #51, opened 2020-09-25
Create Oneshot CI workflow

bwoodsend:
Following on from my [one-off CI](https://github.com/bwoodsend/pyinstaller-hooks-contrib/actions/runs/227584385) to test @rokm's tensorflow hooks, I've generalised that workflow to a manually triggered, parametrized action so that we can perform these tests for any library.

This PR is made up of 5 commits. The main ones are 24b7d61, which adds the workflow itself, and ea1df88, which adds a CLI script to conveniently interact with it using the Github API. I'd rather these commits weren't squashed on merge so I'm PR-ing them so you can review but I'll just push to master once we're happy.

**Usage through the web UI:**
- Go to Actions:
image
- Oneshot test:
image
- Run workflow:
image
- It'll prompt you for parameters. Fill them in:
image

**Usage through the CLI**

If you're running repeatedly, the web UI gets annoying. Instead use the CLI. Usage is documented in its `--help` [here](https://github.com/pyinstaller/pyinstaller-hooks-contrib/blob/24b7d619a8c5a27f993a7f1e968221e0eb78f0e6/scripts/cloud-test.py#L112).

-----

Some [sample output](https://github.com/bwoodsend/pyinstaller-hooks-contrib/actions/runs/272262147).

We (having write access) to this repo can trigger this action on this repo. Or any contributor will be able to use it on their own fork.


bwoodsend:
@rokm For some reason I can't add you as a reviewer (guess we need to add you some group somewhere). But consider your review requested...

bwoodsend:
Ughh, damm. I thought I'd dealt with the linter.

Legorooj (approved):
Awesome work!

scripts/cloud-test.py
```diff
@@ -0,0 +1,213 @@
...
+
+
+def _get_current_branch():
+    """Get the branch currently active locally in git.
+
+    The location of this repo is defined based on the location of this script so it is current-working-dir
+    independent.
+    """
+    branch_command = "git -C \"{}\" branch".format(HERE)
+    branch_output = subprocess.check_output(branch_command).decode()
```
rokm, line 59: This does not work on linux - it raises `FileNotFoundError`.

Either `shell=True` needs to be added, or `branch_command` needs to be in the argument-list form (which is probably the preferable approach).

rokm (changes requested):

rokm:
This is awesome indeed! 

Aside from the error while trying to determine branch, it looks good to me. 

Perhaps the CLI script could be made executable for linux and macOS users (it would need the shebang at the top, and the file would need to be marked as executable), but that can be done later...

bwoodsend:
> the file would need to be marked as executable

Would I need to `sudo chmod`-ify it then add those permissions into git? I have a vague memory of having to do something similar with a `.sh` script.

scripts/cloud-test.py
bwoodsend: OK. I had a weird issue in another project where a mac CI stubbornly refused to accept the argument-list form which is why I've been wary of it here. I'll stick with adding `shell=True` unless you're sure mac will be ok with the argument-list form?

rokm:
> > the file would need to be marked as executable
> 
> Would I need to `sudo chmod`-ify it then add those permissions into git? I have a vague memory of having to do something similar with a `.sh` script.

On linux, doing `chmod +x scripts/cloud-test.py` and then `git commit -a` (or add and commit) would do the trick. You don't need root privileges for that, as the script file is owned by you (= the regular user).

rokm:
> > > the file would need to be marked as executable
> > 
> > 
> > Would I need to `sudo chmod`-ify it then add those permissions into git? I have a vague memory of having to do something similar with a `.sh` script.
> 
> On linux, doing `chmod +x scripts/cloud-test.py` and then `git commit -a` (or add and commit) would do the trick. You don't need root privileges for that, as the script file is owned by you (= the regular user).

Ah, apparently you can do it directly with git, too: `git update-index --chmod=+x <file>`.

scripts/cloud-test.py
rokm: On macOS `FileNotFoundError` is raised as well, but the argument-list form seems to work for me.

rokm: It also saves you the trouble of quoting the path (and, if I recall correctly, different escaping rules between platforms/shell interpreters).

bwoodsend:
OK. I think that resolves both your points...

rokm (approved):

rokm:
It does, thanks!

PR #52, opened 2020-09-29
Add hooks for skimage.feature and skimage.graph

rokm:
This PR adds a basic import test for `skimage` based on the issues encountered while solving pyinstaller/pyinstaller#5197. 

Based on the test failures under scikit-image 0.17.2, we need additional hooks for `skimage.feature` and `skimage.graph` to deal with the missing (sub)modules.

### DIFF
--- a/src/_pyinstaller_hooks_contrib/tests/test_libraries.py
+++ b/src/_pyinstaller_hooks_contrib/tests/test_libraries.py
@@ -361,7 +361,7 @@ def test_pydivert(pyi_builder):
 
 @importorskip('skimage')
 @pytest.mark.skipif(not is_module_satisfies('skimage >= 0.16.0'),
-    reason='The test supports only skimage 0.16.0 or newer.')
+                    reason='The test supports only skimage 0.16.0 or newer.')
 @pytest.mark.parametrize('submodule', [
     'color', 'data', 'draw', 'exposure', 'feature', 'filters', 'future',
     'graph', 'io', 'measure', 'metrics', 'morphology', 'registration',
--- a/src/_pyinstaller_hooks_contrib/tests/test_libraries.py
+++ b/src/_pyinstaller_hooks_contrib/tests/test_libraries.py
@@ -14,6 +14,7 @@ import pytest
 
 from PyInstaller.compat import is_darwin
 from PyInstaller.utils.tests import importorskip, xfail, skipif_win
+from PyInstaller.utils.hooks import is_module_satisfies
 
 
 @importorskip('jinxed')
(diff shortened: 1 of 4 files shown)


src/_pyinstaller_hooks_contrib/tests/test_libraries.py
```diff
@@ -356,3 +357,20 @@ def test_pydivert(pyi_builder):
...
+def test_skimage(pyi_builder):
+    # Explicitly import all skimage sub-packages.
+    # The list below is valid for skimage >= 0.16.0
+    subpackages = ['color', 'data', 'draw', 'exposure', 'feature',
+                   'filters', 'future', 'graph', 'io', 'measure',
+                   'metrics', 'morphology', 'registration',
+                   'restoration', 'segmentation', 'transform',
+                   'util', 'viewer']
+
+    code = "\n".join(["import skimage.{0}".format(x) for x in subpackages])
```
bwoodsend, line 374: I'd rather run each of these as separate tests per submodule. That way we'd know of a hidden dependency of one is being masked because another submodule which also requires the hidden dependency. but either does it explicitly or has a hook.

bwoodsend: Probably would take a while to run though...

rokm: I initially had separate tests for each submodule, but all 36 of them took a while to run, so I figured a combined test (effectively 2 tests) might be a better idea. But I agree that separate tests would be better indicators for issues, so I can split them up again, if we don't mind the prolonged test time.

bwoodsend: Hmm, well I certainly don't want all 36 of them running as part of our main CI. Personally, I'd rather we split them but either omit `scikit-image` from our requirements file or create a `@too_slow_for_CI` marker to skip them for our regular per-push CI. And then just test the slow test once and forget about it. @Legorooj What do you think?

Legorooj: Remove it from the requirements, and use the `pytest.parameterize` decorator to parameterize the test. We can maybe schedule a build or something that runs _every_ test but we can look at that later.

rokm: OK, I put the submodules into `pytest.parametrize` again, so we have a separate test for each one. And I couldn't find `scikit-image` anywhere in requirements, so it shouldn't be slowing the tests by default.

bwoodsend (approved):
Well, if it works, I'm happy.

PR #54, opened 2020-10-04
Exclude tinycss2 hook for versions after 1.0.0

danyeaw:
tinycss2 https://github.com/Kozea/tinycss2/pull/27 updated the packaging which removes the need for a hook.

### DIFF
--- a/news/54.update.rst
+++ b/news/54.update.rst
@@ -1,0 +1,1 @@
+Change hook for ``tinycss2``, no longer needed after version 1.0.0.
--- a/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-tinycss2.py
+++ b/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-tinycss2.py
@@ -12,11 +12,12 @@
 
 
 """
-Hook for tinycsss2. tinycss2 is a low-level CSS parser and generator.
+Hook for tinycss2. tinycss2 is a low-level CSS parser and generator.
 https://github.com/Kozea/tinycss2
 """
 from PyInstaller.utils.hooks import collect_data_files
 
 
+# Hook no longer required for tinycss2 >= 1.0.0
 def hook(hook_api):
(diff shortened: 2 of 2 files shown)


bwoodsend:
I take it earlier versions that did need the hook are old enough to be considered obsolete? i.e There's no real need to wrap the old hook in a `if not is_module_satisfies("tinycss>=x.y.z"):`?

danyeaw:
@bwoodsend Thanks, I think I was going with the thought that the new version is out and projects should update. Keeping the hook around is more conservative, but also increases the burden for this project. Any thoughts on how long to keep the hook around in general after an update?

danyeaw:
@bwoodsend I updated this PR to only exclude the hook for versions after 1.0.0. 👍

bwoodsend:
Hmm, well quite often not upgrading packages is the only way to get PyInstaller to behave itself. The latest matplotlib for example brought in changes which broke PyInstaller's hook for it. So I'd rather not jettison support just because you can upgrade your way around it in this case. 

I think the status quo is just to keep the hooks until they either start getting in the way or they break something. AFAIK thinycss2 isn't causing us issues and we don't have any tests for tinycss2 so it doesn't even slow our CI. So I'd rather hang onto it for now. Maybe we could add an `# XXX: This hook is redundant since [version]` for now. @Legorooj Do you have any thoughts on this?

Legorooj:
I'd prefer just a note saying `# This hook isn't required after tinycss2 1.0.0`, rather than changing the hook.

danyeaw:
@Legorooj You want me to remove `is_module_satisfies` check and replace it with a comment?

Legorooj:
@danyeaw yes please.

danyeaw:
@Legorooj Ok, updates made :+1:

bwoodsend (approved):

Legorooj (approved):
Thanks @danyeaw

PR #57, opened 2020-10-11
hooks: Add hook for pydantic.

htgoebel:
pydanic is available in two flavours: a pure-python implementation and a
compiled implementation. This hook adds hidden imports required by the later
one.

This hook also requires https://github.com/pyinstaller/pyinstaller/pull/5157 to be merged.


bwoodsend:
Superseeded by #78.

PR #58, opened 2020-10-14
hooks: add a hook for win32ctypes.core

rokm:
We need a hook to collect all submodules from `win32ctypes.core.cffi` or `win32ctypes.core.ctypes` (depending on availability of `cffi`) to avoid missing module errors when using `win32ctypes` pywin32 compatibility layer.


bwoodsend (approved):
Looks good to me.

Legorooj (approved):

PR #59, opened 2020-10-21
hooks: zmq: avoid triggering compilation of C extension for cffi backend

rokm:
This PR aims to work around the `UnicodeDecodeError` issue from pyinstaller/pyinstaller#2258. Trying to collect all submodules of `zmq.backend` triggers compilation of C extension for the `cffi` backend (provided cffi and C compiler are installed), but individually collecting each backend submodule does not, which means no potentially breaking error messages from MSC on Windows.

After some more investigation, it turned out that cffi backend requires collection of data files; otherwise lack thereof triggers a file-not-found error when trying to import the backend module.

But even with that, the `cffi` backend does not seem to be usable in a frozen program, as its import fails with `Failed to import zm.backend.cffi: PyZMQ CFFI backend couldn't find zeromq: invalid command 'build`. The error originates from `cffi`/`distutils` itself, so perhaps something is broken there. (And even if that worked, ZeroMQ development files would need to be installed on the target machine). So in short, I suspect `cffi` backend never really worked, so there's little point in collecting it at all. I put the corresponding statements under `if False` as a sort of documentation, and in case anyone wants to try making it work.

### DIFF
--- a/news/59.update.rst
+++ b/news/59.update.rst
@@ -1,0 +1,1 @@
+Avoid collecting non-functional ``zmq.backend.cffi`` backend in the ``zmq`` hook, and thus also prevent an attempt at compilation of its C extension during module collection. 
--- a/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-zmq.py
+++ b/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-zmq.py
@@ -15,7 +15,17 @@
 Hook for PyZMQ. Cython based Python bindings for messaging library ZeroMQ.
 http://www.zeromq.org/
 """
-from PyInstaller.utils.hooks import collect_submodules, get_module_file_attribute
-from PyInstaller.compat import is_win
+from PyInstaller.utils.hooks import collect_submodules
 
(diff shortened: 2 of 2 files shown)


bwoodsend:
To be honest, expecting dynamically generated and compiled C to be portable is up there with the glass hammer or the chocolate kettle in terms of feasiblity. Even if we did get working once I'd be reluctant to keep the hook for fear that we'd waste too much time maintaining it. 

I don't suppose there's any way to detect which backend the user is likely to use so we can warn them?

rokm:
It seems that usually the preferred backend is `cython`, except for PyPy python:
https://github.com/zeromq/pyzmq/blob/bf3a79eee23da0dfa36cc12f4841e1b9a5028b6a/zmq/backend/__init__.py#L13-L40

And even in that case, any exception during import will have it fall back to `cython` again.

There is no fallback if user explicitly defines backend via PYZMQ_BACKEND at runtime, but I imagine that PYZMQ_BACKEND=cffi never worked with pyinstaller anyway...

bwoodsend (approved):

Legorooj (approved):
This'll do, thanks. May I suggest you send it over to ZMQ as well so they have it in their codebase? We can remove ours in a bit that way.

PR #60, opened 2020-10-26
 hooks: add hook for metpy 

rokm:
Partially fixes pyinstaller/pyinstaller#5247 (missing metadata and data files). Full fix requires pyinstaller/pyinstaller#5284, which is also required for the added test to pass.

### DIFF
--- a/news/60.new.rst
+++ b/news/60.new.rst
@@ -1,0 +1,1 @@
+Add hook for ``metpy``.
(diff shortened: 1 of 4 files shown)


rokm:
This should be ready now. I was trying to pin latest version of MetPy (1.0), but it turns out they switched from `pkg_resources` to `importlib.resources` / `importlib_resources`, and `importlib_resources` does not play nicely with PyInstaller...

rokm:
Oneshot test: https://github.com/rokm/pyinstaller-hooks-contrib/actions/runs/1221186892

(the two failures on python 3.9 are due to lack of wheels for pinned requirement versions).

rokm:
We now support both 0.12.x series that uses `pkg_resources` and 1.x series that switched to `importlib.resources` / `importlib_resources`.

bwoodsend (approved):

PR #61, opened 2020-10-26
hooks: add hook for branca

rokm:
Partially fixes pyinstaller/pyinstaller#5256 and pyinstaller/pyinstaller#5262 (missing data files for `branca`, which `folium` depends on). Full fix requires pyinstaller/pyinstaller#5284, which is also required for the added test to pass.


rokm:
Superseded by #318.

PR #63, opened 2020-10-30
hooks: update sentry_sdk

eeriks:
Updated `sentry_sdk` hidden imports - pyinstaller stopped working after sentry_sdk 0.19.1


BoboTiG:
Hm that would add a lot of potential unused files. In fact, new errors are coming because they enabled by default `auto_enabling_integrations`. To fix that, one should set `auto_enabling_integrations=False` when initializing the SDK.

I would not be really happy to see all those files packaged if I do not use them. And in the other hand, it will break everyone using an up-to-date version of Sentry.

Someone else should help deciding what is best to :)

eeriks:
Oh.. haven't checked sentry docs since 0.16, didn't know about `auto_enabling_integrations`.

Yes, these changes then are not needed :)

src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-sentry_sdk.py
```diff
@@ -17,4 +17,14 @@
...
+                 "sentry_sdk.integrations.django",
+                 "sentry_sdk.integrations.flask",
+                 "sentry_sdk.integrations.bottle",
+                 "sentry_sdk.integrations.falcon",
+                 "sentry_sdk.integrations.sanic",
+                 "sentry_sdk.integrations.celery",
+                 "sentry_sdk.integrations.rq",
+                 "sentry_sdk.integrations.aiohttp",
+                 "sentry_sdk.integrations.tornado",
+                 "sentry_sdk.integrations.sqlalchemy"]
```
bwoodsend, line 30: Normally we'd use:
```python
from PyInstaller.utils.hooks import collect_submodules
hiddenimports = collect_submodules("sentry_sdk.integrations")
```
instead of big lists of submodules although I notice the original didn't do that either. Did you try `collect_submodules()`? Does it not work?

bwoodsend:
@eeriks Is it possible to use a particular integration without importing it explicitly? i.e. if you were to use the `django` integration, would you expect to have `import django` somewhere else in your code?

src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-sentry_sdk.py
eeriks: Just tried. Doesn't work.
First there is a python error while packaging (`ModuleNotFoundError: No module named 'django'`),but continues and when running executable `ModuleNotFoundError: No module named 'sentry_sdk.integrations.stdlib'
[22269] Failed to execute script __main__`

SimonIT:
Hey guys, I just opened a PR for sentry and then recognized this one. In my PR #71 I add the default enabled integrations dynamically. What do you think about it?

bwoodsend:
I suppose that's a sensible compromise (although I still don't like it). I take it that you can customize by adding integrations with `--hiddenimport=sentry_sdk.integrations.xxx` and removing with `--exludes=sentry_sdk.integrations.xxx`? I don't see how we could ever have completely seamless/automatic support for this library - optionally dependencies are bad for PyInstaller and this is obviously an extreme case of them.

PR #64, opened 2020-10-31
hooks: Add hook for ijson

Steelsmasher:
Hook for the ijson package at: https://github.com/ICRAR/ijson

Ijson is an iterative JSON parser which is useful for reading very large json files. The backend modules necessary to run ijson are dynamically loaded hence the need for this hook.

### DIFF
--- a/news/64.new.rst
+++ b/news/64.new.rst
@@ -1,0 +1,1 @@
+Add hook for ijson which has dynamically loaded backends.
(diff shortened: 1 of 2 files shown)


bwoodsend (changes requested):
This needs a [news entry](https://github.com/pyinstaller/pyinstaller-hooks-contrib/blob/master/news/README.txt) (something like *Add hook for ``ijson`` which has dynamically loaded backends.*) but other than that you're good to go. Don't worry about the failing tests - they're failing due to something else...

bwoodsend (approved):

Legorooj (approved):

PR #65, opened 2020-10-31
Backport changes from pyinstaller/pyinstaller#5291

BoboTiG:
New pytest markers were introduced in https://github.com/pyinstaller/pyinstaller/pull/5291:

 - `@pytest.mark.darwin` (only runs on macOS)
 - `@pytest.mark.linux` (only runs on GNU/Linux)
 - `@pytest.mark.win32` (only runs  on Windows)


bwoodsend:
That was quick. Nice one.

BoboTiG:
I just realized I did not add the NEWs entry. Do you think it is needed?

bwoodsend:
No, it doesn't affect end users so it doesn't get one.

PR #66, opened 2020-11-04
hooks: Add hook for lxml

doudz:
lxml has hiddenimports and is not fully embedded when adding it to hiddenimports

`hiddenimports = ['lxml'] `is not enough
see https://github.com/pyinstaller/pyinstaller/issues/5306

### DIFF
--- a/news/66.new.rst
+++ b/news/66.new.rst
@@ -1,0 +1,1 @@
+Add hook for lxml which has hidden imports.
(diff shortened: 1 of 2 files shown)


bwoodsend (approved):

bwoodsend:
Perfect thanks. We aim to release these monthly so in a few weeks you'll be able to 
```
pip install -U pyinstaller-hooks-contrib
```
But until then use
```
pip install -U https://github.com/pyinstaller/pyinstaller-hooks-contrib/archive/master.zip
```

PR #67, opened 2020-11-09
hooks: win32com: avoid printing ModuleNotFindError when module is unavailable

rokm:
Use a custom `exec_statement()` to query the `win32com.__file__` instead of using `get_module_file_attribute()`, as the latter
will print a `ModuleNotFindError` to stderr when the module is unavailable. The error message, while harmless, appears
to be bothering the users.

Fixes pyinstaller/pyinstaller#3733.

### DIFF
--- a/src/_pyinstaller_hooks_contrib/hooks/pre_safe_import_module/hook-win32com.py
+++ b/src/_pyinstaller_hooks_contrib/hooks/pre_safe_import_module/hook-win32com.py
@@ -22,14 +22,25 @@ module 'win32com.shell' is in reality 'win32comext.shell'.
 
 import os
 
-from PyInstaller.utils.hooks import logger, get_module_file_attribute
+from PyInstaller.utils.hooks import logger, exec_statement
 from PyInstaller.compat import is_win, is_cygwin
 
 
 def pre_safe_import_module(api):
     if not (is_win or is_cygwin):
         return
-    win32com_dir = os.path.dirname(get_module_file_attribute('win32com'))
+    win32com_file = exec_statement(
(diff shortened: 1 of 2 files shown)


src/_pyinstaller_hooks_contrib/hooks/pre_safe_import_module/hook-win32com.py
```diff
@@ -22,13 +22,25 @@
...
+        try:
+            from win32com import __file__
+            print(__file__)
+        except Exception:
+            pass
+        """).strip()
+    if not is_available:
+        logger.debug('win32com: module not not available')
+        return  # win32com unavailable
     win32com_dir = os.path.dirname(get_module_file_attribute('win32com'))
```
bwoodsend, line 44: Is there a reason you chose to keep the `get_module_file_attribute()` rather than using the output from the `exec_statement()` call?

rokm: Ah, I didn't pay enough attention to what `get_module_file_attribute` actually does. But you're right, it is redundant now.

bwoodsend (approved):
That makes more sense :)

bwoodsend:
I'll let @Legorooj merge this when he wake up.

Legorooj (approved):

bwoodsend:
Appveyor is failing which is unfortunate because it's the one running Windows which we're trying to test. But the tests that are failing are `pyenchant` and `pylint` which I've seen fail elsewhere...

PR #69, opened 2020-11-16
hooks: Add hook for afmformats.

paulmueller:
afmformats is a library for opening common atomic force microscopy (AFM) data files.
https://afmformats.readthedocs.io


bwoodsend:
Look good! Thanks a lot.

PR #70, opened 2020-11-17
hooks: update pyproj hook to improve compatibility across different versions

rokm:
On versions prior to 2.3.0, we require a hidden import of `pyproj._datadir`.

On versions 3.0.0 or newer, we require a hidden import of `pyproj._compat`, and on linux/macOS also `distutils.util`.

This allows the test to pass under wide range of pyproj versions, from 2.1.3 to 3.0.0.

### DIFF
--- a/news/70.update.rst
+++ b/news/70.update.rst
@@ -1,0 +1,1 @@
+Update ``pyproj`` hook to improve compatibility across different versions of ``pyproj`` (from 2.1.3 to 3.0.0).
--- a/requirements-test-libraries.txt
+++ b/requirements-test-libraries.txt
@@ -14,6 +14,7 @@ pycryptodome==3.9.7
 pycryptodomex==3.9.7
 pyexcelerate==0.8.0
 pylint==2.4.4
+pyproj==3.0.0
 pyusb==1.0.2
 pyzmq==19.0.0
 Unidecode==1.1.1
--- a/requirements-test-libraries.txt
+++ b/requirements-test-libraries.txt
@@ -14,7 +14,6 @@ pycryptodome==3.9.7
 pycryptodomex==3.9.7
 pyexcelerate==0.8.0
 pylint==2.4.4
-pyproj==3.0.0
 pyusb==1.0.2
 pyzmq==19.0.0
 Unidecode==1.1.1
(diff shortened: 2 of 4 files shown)


rokm:
Full test: https://github.com/rokm/pyinstaller-hooks-contrib/actions/runs/368008739

rokm:
Hm, what should the testing strategy be here (for the CI)? Should we always test the latest available version (i.e., no version specified in the `requirements-test-libraries.txt), a fixed version (with potentially different versions depending on what's available on each python version), or not at all (i.e., leave the library out of `requirements-test-libraries.txt` and test with oneshot tests on-demand)?

bwoodsend:
I think this is something we ought to discuss generally for this repo.

Unless a PR updates a hook we have a test for then our current CI with the package versions frozen tests that neither PyInstaller's `HEAD` nor CI's docker images have changed in a breaking way - neither of which depend on PRs to this repo so I think this kind of test should be done periodically instead of per PR.

Stripping the versions from the requirements files would be useful in that it will catch breaking changes in updated libraries but again I think this should be done periodically instead of per PR. Otherwise we'll be getting failures that incorrectly blame a PR's changes for the fault.

So I personally think the solution is to make our current PR CI periodic (both with and without the version constraints), then encourage the one-shot testing for PRs. One thing we'd have to be careful of is that on updating a hook for a given package, any other packages which depend on the 1st package also get one-shot tested. 

What do you two think?

rokm:
I agree, periodic checks with fixed or latest versions of the libraries (or both) sound like the best approach. This way, we can get an advance warning that something will need fixing. 

The PRs that have corresponding tests available should probably always be checked with one-shot tests across several versions of the library (ideally the ones that are still supported and can be encountered in the wild, although it's sometimes difficult to draw the line). What we would want to avoid are fixes that address current/latest version of the library, but accidentally break the previous ones that might be still in use - and the current CI testing is insufficient for that.

rokm:
I'm dropping `pyproj` from `requirements-test-libraries.txt` due to the latest version not always being available, and/or due to wheels not being always available (which triggers error during build from source due to `proj` itself not being available)...

bwoodsend (approved):

Legorooj (approved):

PR #71, opened 2020-11-18
Add default enabled sentry integrations dynamically to hidden imports

SimonIT:
Sentry added in PR getsentry/sentry-python#625 by default enabled integrations. We need to add them to the hidden imports to be able to launch a application with sentry.

I added the hasattr check for backwards compatibility

### DIFF
--- a/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-sentry_sdk.py
+++ b/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-sentry_sdk.py
@@ -37,4 +37,3 @@ if hasattr(si, '_AUTO_ENABLING_INTEGRATIONS'):
 """
 
 hiddenimports.extend(json.loads(exec_statement(statement)))
-
--- a/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-sentry_sdk.py
+++ b/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-sentry_sdk.py
@@ -27,7 +27,7 @@ import sentry_sdk.integrations as si
 
 integrations = []
 if hasattr(si, '_AUTO_ENABLING_INTEGRATIONS'):
-    # _AUTO_ENABLING_INTEGRATIONS is a list of strings with default enabled integrations 
+    # _AUTO_ENABLING_INTEGRATIONS is a list of strings with default enabled integrations
     # https://github.com/getsentry/sentry-python/blob/c6b6f2086b58ffc674df5c25a600b8a615079fb5/sentry_sdk/integrations/__init__.py#L54-L66
 
     def make_integration_name(integration_name: str):
--- a/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-sentry_sdk.py
+++ b/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-sentry_sdk.py
@@ -22,6 +22,6 @@ hiddenimports = ["sentry_sdk.integrations.stdlib",
 
 if hasattr(si, '_AUTO_ENABLING_INTEGRATIONS'):
     def make_integration_name(integration_name: str):
-        return ".".join(integration_name.split(".")[:-1])
+        return integration_name.rsplit(".", maxsplit=1)[0]
 
     hiddenimports.extend(map(make_integration_name, si._AUTO_ENABLING_INTEGRATIONS))
--- a/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-sentry_sdk.py
+++ b/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-sentry_sdk.py
@@ -9,7 +9,8 @@
 #
 # SPDX-License-Identifier: GPL-2.0-or-later
 # ------------------------------------------------------------------
-import sentry_sdk.integrations as si
+import json
+from PyInstaller.utils.hooks import exec_statement
 
 hiddenimports = ["sentry_sdk.integrations.stdlib",
                  "sentry_sdk.integrations.excepthook",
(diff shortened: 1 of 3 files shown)


src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-sentry_sdk.py
```diff
@@ -9,6 +9,7 @@
 #
 # SPDX-License-Identifier: GPL-2.0-or-later
 # ------------------------------------------------------------------
+import sentry_sdk.integrations as si
```
bwoodsend, line 12: Whenever you import a 3rd party library in a hook you should put it inside an `exec_statement()` ([docs](https://pyinstaller.readthedocs.io/en/stable/hooks.html#useful-items-in-pyinstaller-utils-hooks)). That would mean moving this down to join the code at the bottom which should go inside the exec-ed statement too.

src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-sentry_sdk.py
```diff
@@ -18,3 +19,9 @@
                  "sentry_sdk.integrations.argv",
                  "sentry_sdk.integrations.logging",
                  "sentry_sdk.integrations.threading"]
+
+if hasattr(si, '_AUTO_ENABLING_INTEGRATIONS'):
```
bwoodsend, line 23: Can we have a comment saying exactly what type/format `si._AUTO_ENABLING_INTEGRATIONS` is? I'm guessing list of string names of integrations where the names exclude the `sentry_sdk.integrations.` prefix?

src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-sentry_sdk.py
```diff
@@ -18,3 +19,9 @@
...
+    def make_integration_name(integration_name: str):
+        return ".".join(integration_name.split(".")[:-1])
```
bwoodsend, line 25: Marginally clearer IMO.
```suggestion
        return integration_name.rsplit(".", maxsplit=1)[0]
```

bwoodsend (commented):
Once you've dealt with my review comments could you launch our CI tester? From the root of your copy of this repo, check your in your `sentry_django_integration` branch, then run:
```shell
python scripts\cloud-test.py --os=windows --os=ubuntu --py=3.6,3.8 --browser sentry-sdk==0.15.1, sentry-sdk==0.19.3
```
You'll need to create a github authentication token if you don't already have one because github have stopped accepting username and password.

src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-sentry_sdk.py
SimonIT: The problem is, if I'm correct, that `exec_statement()` communicates via prints. If I print the list, I would have to convert it back to list to extend the `hiddenimports` list which would add a lot of overhead. I would have to encode the list as json, print the json and then to parse the json for getting the original list.

src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-sentry_sdk.py
SimonIT: Yes, that's the list https://github.com/getsentry/sentry-python/blob/c6b6f2086b58ffc674df5c25a600b8a615079fb5/sentry_sdk/integrations/__init__.py#L54-L66

src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-sentry_sdk.py
bwoodsend: That list is tiny. I get 14 microseconds of overhead...

src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-sentry_sdk.py
bwoodsend: Can you put that url into the code.

SimonIT:
@bwoodsend How should I handle the trailing whitespace problem as reported by the linter? Or does python inside `exec_statement` need no indentation?

rokm:
> @bwoodsend How should I handle the trailing whitespace problem as reported by the linter? Or does python inside `exec_statement` need no indentation?

Linter is complaining about the extra whitespace at the end of the line 30 (after "integrations").

src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-sentry_sdk.py
```diff
@@ -18,3 +20,21 @@
...
                  "sentry_sdk.integrations.logging",
                  "sentry_sdk.integrations.threading"]
+
+statement = """
+import json
+import sentry_sdk.integrations as si
+
+integrations = []
+if hasattr(si, '_AUTO_ENABLING_INTEGRATIONS'):
+    # _AUTO_ENABLING_INTEGRATIONS is a list of strings with default enabled integrations
```
bwoodsend, line 30: I think you're confusing *trailing* whitespace with *leading* whitespace (a.k.a indentation). All the linter is moaning about is (as @rokm says):
```suggestion
    # _AUTO_ENABLING_INTEGRATIONS is a list of strings with default enabled integrations
```

bwoodsend:
Looks good.

> Once you've dealt with my review comments could you launch our CI tester? From the root of your copy of this repo, check into your `sentry_django_integration` branch, then run:
> 
> ```shell
> python scripts\cloud-test.py --os=windows --os=ubuntu --py=3.6,3.8 --browser sentry-sdk==0.15.1, sentry-sdk==0.19.3
> ```
> 
> You'll need to create a github authentication token if you don't already have one because github have stopped accepting username and password.

Can you go ahead and do this now? If it passes then we should be ready to merge this.

SimonIT:
I did it already
https://github.com/SimonIT/pyinstaller-hooks-contrib/actions/runs/372832011

bwoodsend:
Perfect - thanks for your contribution!

PR #77, opened 2020-11-27
Add prettytable hook

krellsebastian:
Add hook-prettytable.py: pyinstaller not working with prettytable module. 
Fixes https://github.com/pyinstaller/pyinstaller/issues/5319

### DIFF
--- a/news/72.new.rst
+++ b/news/72.new.rst
@@ -1,0 +1,1 @@
+Add a hook for prettytable which requires its distribution metadata.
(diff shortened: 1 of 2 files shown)


bwoodsend (approved):
Nailed it in one! Perfect thanks.

news/72.new.rst
```diff
@@ -0,0 +1 @@
+Add a hook for prettytable which requires its distribution metadata.
```
rokm, line 1:
```suggestion
Add a hook for ``prettytable`` which requires its distribution metadata.
```

rokm:
The number in news fragment filename should match the number of this PR (i.e., `77.new.rst` instead of `72.new.rst`). Other than that, look good. Thanks!

PR #78, opened 2020-12-05
Add hook for pydantic

rokm:
The PyPi version of `pydantic` comes with all its modules compiled as cpython extensions. This prevents `pyinstaller` from properly discovering the dependant modules, both from the `pydantic` package itself and from the standard library.

Fixes pyinstaller/pyinstaller#5359.

### DIFF
--- a/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-pydantic.py
+++ b/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-pydantic.py
@@ -10,24 +10,15 @@
 # SPDX-License-Identifier: GPL-2.0-or-later
 # ------------------------------------------------------------------
 
-from PyInstaller.utils.hooks import exec_statement, collect_submodules
+from PyInstaller.utils.hooks import get_module_attribute, collect_submodules
 
 # By default, pydantic from PyPi comes with all modules compiled as
-# cpython extensions, which seems to  prevent the pyinstaller from
-# automatically picking the submodules
-is_compiled = exec_statement(
-    """
-        import pydantic
-        if pydantic.compiled:
-            print('\\nTrue')
-        else:
-            print('\\nFalse')
-    """
-).split()[-1] == 'True'
-
+# cpython extensions, which seems to prevent pyinstaller from automatically
+# picking up the submodules
+is_compiled = get_module_attribute('pydantic', 'compiled') == 'True'
 if is_compiled:
     # Compiled version; we need to manually collect the submodules from
-    # pyadantic...
+    # pydantic...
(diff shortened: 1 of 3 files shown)


rokm:
Here are test matrices for [before](https://github.com/rokm/pyinstaller-hooks-contrib/actions/runs/402917985) and [after](https://github.com/rokm/pyinstaller-hooks-contrib/actions/runs/402946595). 

I suspect that the versions that were passing even before the changes are not compiled.

Older versions of `pydantic` on linux seem to be still failing with the same symptoms as before - I've verified that those are extension-compiled versions. And I suspect that [this import](https://github.com/samuelcolvin/pydantic/blob/e65d11240a6febcd569ad44a61e2374eb00b9ba6/pydantic/dataclasses.py#L1) is to blame. It is gone from `pydantic` 1.4 and later, so those work fine.

src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-pydantic.py
```diff
@@ -0,0 +1,40 @@
...
+# automatically picking the submodules
+is_compiled = exec_statement(
+    """
+        import pydantic
+        if pydantic.compiled:
+            print('\\nTrue')
+        else:
+            print('\\nFalse')
+    """
+).split()[-1] == 'True'
```
bwoodsend, line 26: There's a `get_module_attribute()` hook utility:
```suggestion
is_compiled = get_module_attribute('pydantic', 'compiled') == 'True'
```
Obviously you'd need to import it.

bwoodsend:
We already have #57 but this looks a lot less scary so I'd definitely go with this one. 

Closes #57.

src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-pydantic.py
rokm: Thanks!

bwoodsend:
I take it then that you wish to drop support for anything below 1.4?

Realistically, I feel that this kind of self-induced mess is for package authors to clear up - not us.

rokm:
> I take it then that you wish to drop support for anything below 1.4?
> 
> Realistically, I feel that this kind of self-induced mess is for package authors to clear up - not us.

I don't think there's anything we can actually do about compiled versions below 1.4. If someone absolutely has to use pyinstaller with those old versions, they'll need to install the non-compiled variant.

bwoodsend (approved):
Ok then.

Legorooj (approved):
I seem to remember there being a discussion about reading the imports from python extension modules somewhere, but can't remember off the top of my head. Will look into it again when I tackle modulegraph in January.

PR #79, opened 2020-12-14
tests: define the darwin/linux/win32 pytest markers

rokm:
In order to avoid `PytestUnknownMarkWarning` warnings.

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,2 +1,7 @@
 [pytest]
 addopts=--maxfail=3
+
+markers =
+    darwin: only run on macOS
+    linux: only runs on GNU/Linux
+    win32: only runs on Windows


bwoodsend:
Hmm, I thought this was all done in  *Backport changes from pyinstaller/pyinstaller#5291* #65. Looking back at #65 I see the changes to the conftest from the main PyInstaller repo weren't applied to this repo. I'd have thought we'd need to do that? Is it really enough to just define those markers without specifying what they mean somewhere?

rokm:
The "meaning" of markers is actually defined [here](https://github.com/pyinstaller/pyinstaller/blob/3051ccb3bdf4a1413ab7d4dad8a8b845aef51593/PyInstaller/utils/conftest.py#L90-L103) - although the message is incorrect, which is how I ended up poking around that in the first place...

So I think it is enough, as long as PyInstaller 4.1+ is used.

rokm:
I.e, the `conftest` from this repo simply imports everything from the main repo's `conftest`.

bwoodsend:
Ahh ok. Makes sense. All our CI uses PyInstaller 4.1 so we should be ok.

bwoodsend (approved):

PR #80, opened 2020-12-15
Add a hook for torchvision.ops

rokm:
Add a hidden import of `torchvision._C` extension module that we fail to pick up automatically. This extension module is required in functions from `torchvision.ops.*` modules, where its absence causes the `_assert_has_ops()` call to fail.
    
Fixes pyinstaller/pyinstaller#5386

### DIFF
--- a/news/80.new.rst
+++ b/news/80.new.rst
@@ -1,0 +1,1 @@
+Add hook for ``torchvision.ops`` to ensure that the required extension module (``torchvision._C``) is collected.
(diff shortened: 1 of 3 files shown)


bwoodsend:
Can you run the CI run tests? Maybe just do one version of Python?

rokm:
Here's the run for the latest stable version under python 3.9 on Windows and Linux: https://github.com/rokm/pyinstaller-hooks-contrib/runs/1557959180?check_suite_focus=true.

bwoodsend (approved):

PR #81, opened 2021-01-01
hooks: pydantic: add missing dataclasses and distutils.version hidden imports

rokm:
Addendum to previous attempt at the `pydantic` hook (#78).

As established in pyinstaller/pyinstaller#5432, the `dataclasses` module from standard library also needs to be added to `hiddenimports`. In older versions of compiled `pydantic`, this missing module would cause an error when importing `pydantic` itself (https://github.com/pyinstaller/pyinstaller-hooks-contrib/pull/78#issuecomment-739382401), while in newer ones the import is done within the functions and so the error is deferred until the function call.

Therefore, this partially addresses [the failing linux test from the previous one-shot test](https://github.com/rokm/pyinstaller-hooks-contrib/actions/runs/402946595). Afterwards, another issue crops up; versions of `pydantic` prior to 1.4 use `distutils.version` in `pydantic.version`, so that needs to be added to hidden imports as well (although judging by the tests, this actually affects only linux - presumably on other platforms, `distutils` is already pulled in by some other dependency).

Now, the pydantic import test passes on all tested versions [on Windows and Linux](https://github.com/rokm/pyinstaller-hooks-contrib/actions/runs/456533325), as well as [on macOS](https://github.com/rokm/pyinstaller-hooks-contrib/actions/runs/456542698).

### DIFF
--- a/news/81.update.rst
+++ b/news/81.update.rst
@@ -1,0 +1,1 @@
+Add missing ``dataclasses`` hidden import to ``pydantic`` hook.
--- a/news/81.update.rst
+++ b/news/81.update.rst
@@ -1,1 +1,3 @@
 Add missing ``dataclasses`` hidden import to ``pydantic`` hook.
+Add missing ``distutils.version`` hidden import to ``pydantic`` hook for
+versions of ``pydantic`` prior to ``1.4``.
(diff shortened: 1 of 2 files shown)


bwoodsend (approved):

Legorooj:
Looking good!

Issue #84, opened 2021-02-01
There is no documentation for running the test suite in news/README.txt


RS357:
There is no documentation for running the test suite.

PR #85, opened 2021-02-01
Added documentation on running test suite.

RS357:
There was no information on running the test suite so it was added
to news/README.txt. Fixes issue #84.

### DIFF
--- a/news/README.txt
+++ b/news/README.txt
@@ -37,3 +37,14 @@ You add the following file into the "news" directory, called "9480.new.rst".
     Add a hook for ``foobar``, which has a hidden import.
 
 Done! The changelog entry has been added, and your PR is ready for review.
+
+
+Test Suite
+-----------
+
+To run the test suite, install dependencies in root:
+* pip3 install -e .
+
+Then run the test suite in root:
+* If using Windows, run: `python -m PyInstaller.utils.run_tests`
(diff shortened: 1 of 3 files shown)


news/README.txt
```diff
@@ -37,3 +37,14 @@ You add the following file into the "news" directory, called "9480.new.rst".
...
+
+Test Suite
+-----------
+
+To run the test suite, install dependencies in root:
+* pip3 install -e .
+
+Then run the test suite in root:
+* If using Windows, run: `python -m PyInstaller.utils.run_tests`
+* If using Linux, run: `python3 -m PyInstaller.utils.run_tests`
```
bwoodsend, line 50: pip follows the same pattern as python in that it's pip on Windows but pip3 everywhere else. macOS also uses python3 and pip3. I believe the reason is because Linux and mac have both come with python 2.7 preinstalled but that's just a guess.

bwoodsend: Also don't we need a `pip install -r requirements-test.txt`?

bwoodsend (changes requested):

news/README.txt
RS357: @Legorooj  anything else to add?

bwoodsend (approved):

Legorooj (approved):
Looking good!

PR #86, opened 2021-02-04
Add hook for the pygraphviz library

vmiklos:
pygraphviz assumes a working graphviz installation.
https://graphviz.org/download/ recommends `brew install graphviz` on
macOS and `choco install graphviz` on Windows, so look for the graphviz
binaries / configs where those tools install by default.

### DIFF
--- a/news/86.new.rst
+++ b/news/86.new.rst
@@ -1,0 +1,1 @@
+Added a hook for the 'pygraphviz' library
--- a/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-pygraphviz.py
+++ b/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-pygraphviz.py
@@ -1,0 +1,41 @@
+# ------------------------------------------------------------------
+# Copyright (c) 2021 PyInstaller Development Team.
+#
+# This file is distributed under the terms of the GNU General Public
+# License (version 2.0 or later).
+#
+# The full license is available in LICENSE.GPL.txt, distributed with
+# this software.
+#
+# SPDX-License-Identifier: GPL-2.0-or-later
+# ------------------------------------------------------------------
+
+import glob
+import os
+import shutil
+
+from PyInstaller.compat import is_win, is_darwin
+
+binaries = []
+datas = []
+
+if is_darwin:
+    # The dot binary in PATH is typically a symlink, handle that.
+    # graphviz_bindir is e.g. /usr/local/Cellar/graphviz/2.46.0/bin
+    graphviz_bindir = os.path.dirname(os.path.realpath(shutil.which("dot")))
+    for binary in glob.glob(graphviz_bindir + "/*"):
+        binaries.append((binary, "."))
+    # graphviz_bindir is e.g. /usr/local/Cellar/graphviz/2.46.0/lib/graphviz
+    graphviz_libdir = os.path.realpath(graphviz_bindir + "/../lib/graphviz")
+    for binary in glob.glob(graphviz_libdir + "/*.dylib"):
+        binaries.append((binary, "graphviz"))
+    for data in glob.glob(graphviz_libdir + "/config*"):
+        datas.append((data, "graphviz"))
+
+if is_win:
+    for binary in glob.glob("c:/Program Files/Graphviz*/bin/*.exe"):
+        binaries.append((binary, "."))
+    for binary in glob.glob("c:/Program Files/Graphviz*/bin/*.dll"):
+        binaries.append((binary, "."))
+    for data in glob.glob("c:/Program Files/Graphviz*/bin/config*"):
+        datas.append((data, "."))


vmiklos:
FWIW, I had this under https://github.com/vmiklos/ged2dot/blob/master/pyi/hook-pygraphviz.py previously, so I tested this on macOS and Windows with ged2dot's pygraphviz usage.

bwoodsend:
Ahh, nice one. Thanks for sharing.

bwoodsend (commented):
Is graphviz's `bin` folder expected to be in `PATH` do you know? It may or may not be safer to use `shutil.which("dot")` and work from there than to hard-code the installation paths.

vmiklos:
> Is graphviz's `bin` folder expected to be in `PATH` do you know? It may or may not be safer to use `shutil.which("dot")` and work from there than to hard-code the installation paths.

This works for macOS, done now. shutil.which() finds /usr/local/bin/dot, which is a symlink, but once resolving that, we can find the grapviz folder. It doesn't work for Windows, which installs a stub binary to PATH, so shutil.which() finds only c:\ProgramData\chocolatey\bin\dot.exe, and os.path.realpath() doesn't help us there.

So I've updated the macOS part, but left Windows unchanged. I also realized that while my ged2dot usage of pygraphviz only wants the "dot" binary, pygraphviz itself wants to also invoke other graphviz binaries as well, so I changed bin/dot[.exe] to bin/*[.exe]. I believe this is safe to do, as both platforms have a bin directory dedicated to graphviz.

bwoodsend:
Ahh, chocolately and its shims...

I had a brief look to see if the Darwin if block could also work for Linux. It does except that the `bin` folder is `/usr/bin` so globbling it takes half the operating system with it. So almost but no - we'll save this for another PR.

news/84.new.rst
```diff
@@ -0,0 +1 @@
+Added a hook for the 'pygraphviz' library
```
bwoodsend, line 1: This is PR #86 - not 84. The file should be named as such.

bwoodsend (changes requested):

news/84.new.rst
vmiklos: Ah, I thought this is just a counter. :-) Fixed now.

bwoodsend (approved):

vmiklos:
Am I expected to do anything here? The "macOS CI / test (3.5)" seems to be broken because:

> ERROR: Package 'pyinstaller' requires a different Python: 3.5.10 not in '>=3.6'

That looks orthogonal to this PR.

bwoodsend:
Oh, ignore CI. It's useless unless you're providing new tests. I'm just waiting for @rokm's review.

rokm (approved):
This looks like a nightmare to collect in any way that could be considered sane (even when operating under greatly relaxed definition of sanity)...

As such, while I dislike the way things are hard-coded and that only Windows and macOS (and even there, only subset of possible configurations) are supported, I think this is a good starting point. 

One thing I wonder, though - do we need to explicitly collect dynamic libraries? I think `cdt`, `cgraph` and `gvc` should be pulled in automatically because the `_graphviz` python wrapper/extension is linked against them. Similarly, since the programs are added as binaries, they will probably go through dependency analysis as well, so all other required libraries should be picked up automatically - or are they not?

I think we could discover the base path of `graphviz` installation by looking for its libraries (e.g., `cdt`) using `ctypes.util`, but that can be left for the next revision of the hook. Also, sooner or later we will probably want to come up with explicit list of executables to collect from graphviz, so we can make it work on linux as well (but that too, is probably beyond the scope of this PR).

bwoodsend:
Thanks @vmiklos. This'll be included in the next release which should be at the end of the month.

vmiklos:
> do we need to explicitly collect dynamic libraries?

My understanding is that there are 2 types of graphviz shared objects: e.g. dot links against libgvc.so.6, and pyinstaller handles that out of the box. And then there are the graphviz plugins, which some of the shared objects open via dlopen(), which pyinstaller can't detect easily. So I think collecting these plugins is needed, e.g. PNG output is not working without them.

> I think we could discover the base path of graphviz installation by looking for its libraries (e.g., cdt) using ctypes.util

I like this idea, but how would that work when all we know (on Windows) is the location of the chocolately shim? In other cases, we already know the install location.

> Also, sooner or later we will probably want to come up with explicit list of executables to collect from graphviz

https://github.com/pygraphviz/pygraphviz/blob/894a45b43aa8ec332758dc170f5a73cbef762a6e/pygraphviz/agraph.py#L1323-L1339

There is already a defined list of executables pygraphviz would look for. Do you think replacing the "*" glob with this list would be an improvement? I dislike copy&paste, but perhaps it's not a big issue, and then the macOS part could be generalized for Linux as well. Or perhaps we could convince the pygraphviz devs to expose this list in a public API, so the hook could take that list from them. :-)

rokm:
> > do we need to explicitly collect dynamic libraries?
> 
> My understanding is that there are 2 types of graphviz shared objects: e.g. dot links against libgvc.so.6, and pyinstaller handles that out of the box. And then there are the graphviz plugins, where some of the shared objects open a plugin via dlopen(), which pyinstaller can't detect easily. So I think collecting these plugins is needed, e.g. PNG output is not working without them.

Ah, good point. But I guess we could glob the plugins via `gvplugin*` (prepending with `lib` if necessary, and adding the OS-specific shared library suffix). Or we could even come up with a rudimentary parser for that `config6` file to get the list of plugins.

And on linux (and probably on macOS as well), plugins already seem to be collected in `lib{,64}/graphviz` subfolder.

> 
> > I think we could discover the base path of graphviz installation by looking for its libraries (e.g., cdt) using ctypes.util
> 
> I like this idea, but how would that work when all we know (on Windows) is the location of the chocolately shim? In other cases, we already know the install location.

The shared libraries that `_graphviz` python extension is linked against (`cdt`, `cgraph`, `gvc`) should be visible in `PATH` regardless of chocolatey shim for the `dot` executable. Otherwise I imagine the extension itself will fail to load. 

> 
> > Also, sooner or later we will probably want to come up with explicit list of executables to collect from graphviz
> 
> https://github.com/pygraphviz/pygraphviz/blob/894a45b43aa8ec332758dc170f5a73cbef762a6e/pygraphviz/agraph.py#L1323-L1339
> 
> There is already a defined list of executables pygraphviz would look for. Do you think replacing the "*" glob with this list would be an improvement? I dislike copy&paste, but perhaps it's not a big issue, and then the macOS part could be generalized for Linux as well. 

Nice. I think checking for executables based on that list would be preferable to the brute-force glob. (And I don't think having our local copy is a big problem).

> Or perhaps we could convince the pygraphviz devs to expose this list in a public API, so the hook could take that list from them. :-)

That would be nice, too. Although that would mean that we'd need to wait for the new version of pygraphviz to be released, and would support only that version or later. Also, I think the long-term goal is to have hooks submitted to their corresponding upstream projects, so they can provide them themselves.

PR #88, opened 2021-02-23
hooks: add hook for pyqtgraph

paulmueller:
pyqtgraph is a fast plotting library https://pyqtgraph.readthedocs.io/en/latest/

### DIFF
--- a/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-pyqtgraph.py
+++ b/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-pyqtgraph.py
@@ -10,7 +10,26 @@
 # SPDX-License-Identifier: GPL-2.0-or-later
 # ------------------------------------------------------------------
 
-from PyInstaller.utils.hooks import collect_data_files
+from PyInstaller.utils.hooks import collect_data_files, collect_submodules
 
-# Data files
-datas = collect_data_files("pyqtgraph", include_py_files=True)
+# include all .ui and image files
+datas = collect_data_files("pyqtgraph",
+                           includes=["**/*.ui", "**/*.png", "**/*.svg"])
+
+# pyqtgraph uses Qt-version-specific templates for the UI elements.
+# There are templates for different versions of PySide and PyQt, e.g.
+#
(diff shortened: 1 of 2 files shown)


src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-pyqtgraph.py
```diff
@@ -0,0 +1,16 @@
...
+# The full license is available in LICENSE.GPL.txt, distributed with
+# this software.
+#
+# SPDX-License-Identifier: GPL-2.0-or-later
+# ------------------------------------------------------------------
+
+from PyInstaller.utils.hooks import collect_data_files
+
+# Data files
+datas = collect_data_files("pyqtgraph", include_py_files=True)
```
bwoodsend, line 16: What's the reasoning behind `include_py_files=True`? It's a pretty unusual thing to do. The only use case I know for it is if Python scripts are used as templates for writing other Python files. If you get `ModuleNotFoundError`s without it then they should usually be hidden imports instead.

paulmueller: OK, I didn't know that. I just thought this would be future-proof, because pyqtgraph is still a moving target. If you insist, I will change this to hidden imports.

paulmueller: I probably should have at least put that in the comment instead of `# Data files`

bwoodsend:
```python
hiddenimports = collect_submodules("pyqtgraph")
```
Would be to future-proof against hidden imports but I wouldn't do that unless `pyqtgraph` are known to do nasty stuff with their imports.

paulmueller: Apparently, it only affects a few Python files which all have "Template" in their name. I updated the hook and elaborated there what I did.

bwoodsend (approved):
Perfect. Very self explanatory.

PR #89, opened 2021-02-26
pygraphviz hook: check for executables based on a list

vmiklos:
Instead of a brute-force glob. This paves the way to get the hook
working on Linux as well, where bundling /usr/bin/* is obviously a
no-go.

See
<https://github.com/pyinstaller/pyinstaller-hooks-contrib/pull/86#issuecomment-774312818>
for more details.

### DIFF
--- a/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-pygraphviz.py
+++ b/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-pygraphviz.py
@@ -19,12 +19,30 @@ from PyInstaller.compat import is_win, is_darwin
 binaries = []
 datas = []
 
+# List of binaries agraph.py may invoke.
+progs = [
+    "neato",
+    "dot",
+    "twopi",
+    "circo",
+    "fdp",
+    "nop",
+    "acyclic",
+    "gvpr",
+    "gvcolor",
+    "ccomps",
+    "sccmap",
+    "tred",
+    "sfdp",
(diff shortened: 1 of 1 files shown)


src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-pygraphviz.py
```diff
@@ -33,8 +51,8 @@
         datas.append((data, "graphviz"))
 
 if is_win:
-    for binary in glob.glob("c:/Program Files/Graphviz*/bin/*.exe"):
-        binaries.append((binary, "."))
+    for binary in progs:
+        binaries.append(("c:/Program Files/Graphviz*/bin/" + binary + ".exe", "."))
```
vmiklos, line 55: This still needs a glob(), let me fix that.

bwoodsend (commented):
Looks good. The `is_darwin` will need to change of course.

vmiklos:
Sure, and also I will have to figure out if the dylib/so difference is something to handle in the hook manually or there is abstraction for that. (Graphviz is somewhat confusing here: as I understand it, macOS plugins loaded via dlopen() have a .so extension, while dynamic libraries loaded by the runtime linker have a .dylib extension. And graphviz have .dylib files, but they are loaded via dlopen()).

In any case, if that's fine for you, I would leave that for a follow-up PR, this one just gets rid of the bin/* glob.

PR #90, opened 2021-03-04
hooks: update pydantic hook for compatibility with v.1.8.0 and later

rokm:
Pydantic 1.8.0 introduced additional dependency on `typing_extensions` module, which needs to be added to `hiddenimports` for the compiled version of pydantic.

Fixes pyinstaller/pyinstaller#5603.

### DIFF
--- a/news/90.update.rst
+++ b/news/90.update.rst
@@ -1,0 +1,1 @@
+Update ``pydantic`` hook for compatibility with v.1.8.0 and later.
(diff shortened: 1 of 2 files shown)


rokm:
Oneshot test: https://github.com/rokm/pyinstaller-hooks-contrib/actions/runs/621651187

bwoodsend (approved):

PR #91, opened 2021-03-04
macOS CI: replace python 3.5 with 3.9

rokm:
Also, disable `fast-fail` so we can get the full picture about failing tests.


bwoodsend:
I have been ignoring that CI for way too long.

PR #93, opened 2021-03-05
pygraphviz hook: add Linux support

vmiklos:
See
<https://github.com/pyinstaller/pyinstaller-hooks-contrib/pull/86#pullrequestreview-584684190>,
this was not done before due to an implicit list of executables, which
is no longer a problem.

Keep the logic as "if windows, else if macos, else", so it can cover not
only Linux, but other similar cases (BSD, etc).

### DIFF
--- a/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-pygraphviz.py
+++ b/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-pygraphviz.py
@@ -15,6 +15,7 @@ import os
 import shutil
 
 from PyInstaller.compat import is_win, is_darwin
+from PyInstaller.depend.bindepend import findLibrary
 
 binaries = []
 datas = []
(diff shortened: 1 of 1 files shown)


src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-pygraphviz.py
```diff
@@ -37,24 +37,27 @@
...
+    for data in glob.glob("c:/Program Files/Graphviz*/bin/config*"):
+        datas.append((data, "."))
+else:
     # The dot binary in PATH is typically a symlink, handle that.
     # graphviz_bindir is e.g. /usr/local/Cellar/graphviz/2.46.0/bin
     graphviz_bindir = os.path.dirname(os.path.realpath(shutil.which("dot")))
     for binary in progs:
         binaries.append((graphviz_bindir + "/" + binary, "."))
     # graphviz_bindir is e.g. /usr/local/Cellar/graphviz/2.46.0/lib/graphviz
     graphviz_libdir = os.path.realpath(graphviz_bindir + "/../lib/graphviz")
```
rokm, line 55: This will need to account for linux systems that use `/usr/lib64/graphviz` on `x86_64`, such as RHEL and Fedora (Ubuntu and Debian do seem to use `/usr/lib/graphviz` regardless of arch, though).

Perhaps discovering the path of `libcdt` would be better approach (if it works on macOS as well?). E.g.,
```python
from PyInstaller.depend.bindepend import findLibrary
graphviz_libdir = os.path.join(os.path.dirname(findLibrary('libcdt')), 'graphviz')
```
returns `'/lib64/graphviz'` on Fedora, because `/lib64` is symlink to `/usr/lib64`.

vmiklos: Good point, I forgot that macOS `lib` doesn't trivially translate to Linux. I'm on openSUSE with the same lib64 naming, will look at this.

vmiklos: I just checked, findLibrary() doesn't work on macOS:

```
>>> graphviz_libdir = os.path.join(os.path.dirname(findLibrary('libcdt')), 'graphviz')
  File "<stdin>", line 1, in <module>
  File ".../lib/python3.9/site-packages/PyInstaller/depend/bindepend.py", line 753, in findLibrary
    assert is_unix, ("Current implementation for Unix only (Linux, Solaris, "
AssertionError: Current implementation for Unix only (Linux, Solaris, AIX, FreeBSD)
```

and given that it's not really needed there, I guess the best is to keep the macOS path as-is and use findLibrary() in the Linux case. I've updated the commit message + the actual code to do this now.

bwoodsend (approved):
Works for me...

rokm (approved):

bwoodsend:
Thanks @vmiklos!

PR #82, opened 2021-01-02
hooks: add hook for googleapiclient.model

rokm:
The `googleapiclient.model` module uses `pkg_resources` and its `get_distribution()` function to query the version of `google-api-python-client` package, therefore we need to collect its metadata.

Fixes pyinstaller/pyinstaller#5332.
Fixes #14.

PR #97, opened 2021-03-11
Added googleapiclient.discovery

sirdeniel:
Implements the fix commented [here](https://github.com/pyinstaller/pyinstaller-hooks-contrib/pull/82#issuecomment-796438598) on #82.
TL;DR: Using google-api-python-client for Blogger API causes `build()` method to raise error and None value due to missing `googleapiclient.discovery` files.

### DIFF
--- a/news/97.update.rst
+++ b/news/97.update.rst
@@ -1,0 +1,2 @@
+Add ```googleapiclient.discovery``` json files to work with services
+like Blogger v3 on the ```build()``` method.
--- a/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-googleapiclient.model.py
+++ b/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-googleapiclient.model.py
@@ -11,8 +11,10 @@
 # ------------------------------------------------------------------
 
 from PyInstaller.utils.hooks import copy_metadata
+from PyInstaller.utils.hooks import collect_data_files
 
 # googleapiclient.model queries the library version via
 # pkg_resources.get_distribution("google-api-python-client").version,
 # so we need to collect that package's metadata
 datas = copy_metadata('google_api_python_client')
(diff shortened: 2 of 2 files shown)


bwoodsend (changes requested):
I seriously doubt you need `include_py_files=True`. What happens if you get rid of it?

Also this needs a [news entry](https://pyinstaller.readthedocs.io/en/stable/development/changelog-entries.html).

rokm:
What's probably needed here is to collect JSON files in `site-packages/googleapiclient/discovery_cache/documents/`, so `include_py_files=True` indeed looks redundant. 

I suppose the data collection should go into separate hook, `hook-googleapiclient.discovery.py` instead of existing `hook-googleapiclient.model.py` to keep things consistent (or alternatively, `hook-googleapiclient.model.py` with proposed changes should be "upgraded" into `hook-googleapiclient.py`)?

sirdeniel:
> I seriously doubt you need `include_py_files=True`. What happens if you get rid of it?
> 
> Also this needs a [news entry](https://pyinstaller.readthedocs.io/en/stable/development/changelog-entries.html).

I'll check this ASAP and comment here. Thanks for the suggestion as I'm noob to this.

sirdeniel:
> What's probably needed here is to collect JSON files in `site-packages/googleapiclient/discovery_cache/documents/`, so `include_py_files=True` indeed looks redundant.
> 
> I suppose the data collection should go into separate hook, `hook-googleapiclient.discovery.py` instead of existing `hook-googleapiclient.model.py` to keep things consistent (or alternatively, `hook-googleapiclient.model.py` with proposed changes should be "upgraded" into `hook-googleapiclient.py`)?

You're right, my apologies for messing up the original script. I'll try my best to get an optimized way to get this done, I'd like some help on the road.

bwoodsend (approved):
Looks good to me.

sirdeniel:
To comment on the behavior, discarding `include_py_files=True` causes to import only the JSON files as suggested by @rokm.

PR #98, opened 2021-03-12
hooks: update zmq hook for compatibility with version 22.0.0+ on Windows

rokm:
Windows wheel of `pyzmq` 22.0.0 introduced separate shared library directory in `site-packages/pyzmq.libs` which is dynamically added to the path and hence eludes our binary analysis.

There's also a `.load_order` file in there that needs to be collected for python versions older than 3.8 (and is also needed on newer
versions to ensure that `pyzmq.libs` directory is created).

### DIFF
--- a/news/98.update.rst
+++ b/news/98.update.rst
@@ -1,0 +1,2 @@
+(Windows) Update ``zmq`` hook for compatibility with new shared libraries 
+location in Windows build of ``pyzmq`` 22.0.0 and later.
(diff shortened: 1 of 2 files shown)


rokm:
Initial tests (damage extent survey): https://github.com/rokm/pyinstaller-hooks-contrib/actions/runs/646283287
Final version (affected platform only): https://github.com/rokm/pyinstaller-hooks-contrib/actions/runs/646553720

bwoodsend (approved):
Hmm, this looks like it'll be a fun library to maintain...

PR #101, opened 2021-03-20
Add hook for pyttsx3.

bwoodsend:
Add hook for `pyttxs3`. Fixes pyinstaller/pyinstaller#3268 and a rather long [nabble thread](https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/pyinstaller/IJXFdvIs--0/D85e6cD7AQAJ).

We do have a hook and test already for `pyttxs` (without the `3`) which is handled similarly but I'd prefer to keep them separate.

There's also a couple of extra commits tucked under here which I can't be bothered to separate into their own PRs. I'll merge this manually to keep the git history.

- 112f500 CI/CD: Oneshot: Allow custom bash install commands. This is so I can run any `sudo apt install [dependencies]` when needed.
- a7fedf0 CI/CD: Oneshot: Remove Python 3.5, add 3.9.

I've done oneshot testing in bits:

- [scripts/cloud-test.py --os=windows --os=macos pyttsx3==2.90, pyttsx3==2.71](https://github.com/bwoodsend/pyinstaller-hooks-contrib/runs/2156860672)
- [scripts/cloud-test.py --os=windows --os=macos --py=all pyttsx3](https://github.com/bwoodsend/pyinstaller-hooks-contrib/actions/runs/671614327)
- [scripts/cloud-test.py --commands="sudo apt install espeak ffmpeg libespeak1" pyttsx3==2.90, pyttsx3==2.71](https://github.com/bwoodsend/pyinstaller-hooks-contrib/actions/runs/671623657)
- [scripts/cloud-test.py --commands="sudo apt install espeak ffmpeg libespeak1" --py=all pyttsx3](https://github.com/bwoodsend/pyinstaller-hooks-contrib/actions/runs/671621133)

They all pass.

### DIFF
--- a/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-pyttsx3.py
+++ b/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-pyttsx3.py
@@ -1,0 +1,30 @@
+# ------------------------------------------------------------------
+# Copyright (c) 2020 PyInstaller Development Team.
+#
+# This file is distributed under the terms of the GNU General Public
+# License (version 2.0 or later).
+#
+# The full license is available in LICENSE.GPL.txt, distributed with
+# this software.
+#
+# SPDX-License-Identifier: GPL-2.0-or-later
+# ------------------------------------------------------------------
+
+# pyttsx3 conditionally imports drivers module based on specific platform.
+# https://github.com/nateshmbhat/pyttsx3/blob/5a19376a94fdef6bfaef8795539e755b1f363fbf/pyttsx3/driver.py#L40-L50
+
+import sys
+
+hiddenimports = ["pyttsx3.drivers", "pyttsx3.drivers.dummy"]
+
+# Take directly from the link above.
+if sys.platform == 'darwin':
+    driverName = 'nsss'
+elif sys.platform == 'win32':
+    driverName = 'sapi5'
+else:
+    driverName = 'espeak'
+# import driver module
+name = 'pyttsx3.drivers.%s' % driverName
+
+hiddenimports.append(name)
--- a/src/_pyinstaller_hooks_contrib/tests/test_libraries.py
+++ b/src/_pyinstaller_hooks_contrib/tests/test_libraries.py
@@ -197,6 +197,14 @@ def test_pyttsx(pyi_builder):
         """)
 
 
+@importorskip('pyttsx3')
+def test_pyttsx3(pyi_builder):
+    pyi_builder.test_source("""
+        import pyttsx3
+        engine = pyttsx3.init()
+    """)
+
+
 @importorskip('pycparser')
 def test_pycparser(pyi_builder):
     pyi_builder.test_script('pyi_lib_pycparser.py')
--- a/.github/workflows/oneshot-test.yml
+++ b/.github/workflows/oneshot-test.yml
@@ -34,6 +34,12 @@ on:
         description: 'Terminate all tests if one fails (true/false).'
         required: true
         default: 'false'
+      commands:
+        description: |
+          Additional installation commands to run from terminal.
+          Ran from bash irregardless of OS.
+          Use ';' to separate multiple commands.
+        required: false
 
 jobs:
   generate-matrix:
@@ -97,6 +103,10 @@ jobs:
         uses: actions/setup-python@v1
         with:
           python-version: ${{ matrix.python-version }}
+
+      - name: Run bash commands
+        if: ${{ github.event.inputs.commands }}
+        run: ${{ github.event.inputs.commands }}
 
       - name: Install dependencies
         run: |
--- a/news/101.new.rst
+++ b/news/101.new.rst
@@ -1,0 +1,1 @@
+Add hook for ``pyttsx3`` whose drivers are hidden imports.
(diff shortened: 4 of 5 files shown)


src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-pyttsx3.py
```diff
@@ -0,0 +1,19 @@
...
+# SPDX-License-Identifier: GPL-2.0-or-later
+# ------------------------------------------------------------------
+
+# pyttsx3 conditionally imports drivers module based on specific platform.
+# https://github.com/nateshmbhat/pyttsx3/blob/5a19376a94fdef6bfaef8795539e755b1f363fbf/pyttsx3/driver.py#L40-L50
+
+
+from PyInstaller.utils.hooks import collect_submodules
+
+hiddenimports = collect_submodules("pyttsx3.drivers")
```
rokm, line 19: Does it make sense to collect all drivers? `sapi5` seems to be Windows-specific, and `nsss` seems macOS-specific. `espeak` should at least nominally be cross-platform, but I guess not so much if it's trying to load `libespeak.so.1` as the shared library... 

So its seems each platform has one actual driver available (in addition to `dummy`).

101.new.rst
```diff
@@ -0,0 +1 @@
+Add hook for ``pyttsx3`` who's drivers are hidden imports.
```
rokm, line 1: whose

src/_pyinstaller_hooks_contrib/tests/test_libraries.py
```diff
@@ -197,6 +197,27 @@ def test_pyttsx(pyi_builder):
...
+    import os
+
+    SOURCE = """
+        import os
+        import pyttsx3
+        engine = pyttsx3.init()
+    """
+    # The above is enough to cause a failure without the hook. It's a lot more
+    # reassuring if it includes the block below too.
+    # But on Windows CI this blocks indefinitely so exclude it in that case.
```
rokm, line 211: I'm not sure if having the test actually speak out the sentence is a good idea. It does ensure that everything works, but does it gracefully handle the case when there's no audio output device present on the system? On the flipside, having it suddenly speak out when running tests on your local PC might also be unexpected...

src/_pyinstaller_hooks_contrib/tests/test_libraries.py
```diff
@@ -197,6 +197,27 @@ def test_pyttsx(pyi_builder):
         """)
 
 
+@importorskip('pyttsx3')
+def test_pyttsx3(pyi_builder):
+    import os
+
+    SOURCE = """
+        import os
```
rokm, line 205: This import looks redundant.

src/_pyinstaller_hooks_contrib/tests/test_libraries.py
```diff
@@ -197,6 +197,27 @@ def test_pyttsx(pyi_builder):
...
+    """
+    # The above is enough to cause a failure without the hook. It's a lot more
+    # reassuring if it includes the block below too.
+    # But on Windows CI this blocks indefinitely so exclude it in that case.
+
+    if not (os.environ.get("CI", False) and os.name == "nt"):
+        SOURCE += """
+        engine.say("I do not like green eggs and ham.")
+        engine.runAndWait()
+        """
```
rokm, line 217: I think SOURCE might look nicer if it contained the conditional block with boolean condition being injected via format(), instead of being appended. But it's probably just a matter of preference...

101.new.rst
bwoodsend: And I put it in the wrong place...

src/_pyinstaller_hooks_contrib/tests/test_libraries.py
bwoodsend: Ok it's gone.

src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-pyttsx3.py
```diff
@@ -0,0 +1,33 @@
...
+#
+# SPDX-License-Identifier: GPL-2.0-or-later
+# ------------------------------------------------------------------
+
+# pyttsx3 conditionally imports drivers module based on specific platform.
+# https://github.com/nateshmbhat/pyttsx3/blob/5a19376a94fdef6bfaef8795539e755b1f363fbf/pyttsx3/driver.py#L40-L50
+
+import sys
+
+from PyInstaller.utils.hooks import collect_submodules
```
rokm, line 18: This unused import would make linter very unhappy.

src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-pyttsx3.py
bwoodsend: Hmm, well the choice of driver is configurable but it doesn't look like a feature that gets used much - or even one which can be used.

src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-pyttsx3.py
bwoodsend: Grrr

rokm (approved):

PR #111, opened 2021-04-21
CI fixup attempt

rokm:


rokm:
Alright, this seems to take care of consistent failures that we're having.

It's mostly re-pinning library versions due to missing wheels and failed attempts to build from source.

In the case of `h5py`, 32-bit wheels are not provided for python 3.9 anymore. I wasn't even aware that Windows CI was running 32-bit version, though... So I've switched it over to 64-bit.

The `dash` test definitely needs timeout larger than former 5 minutes, as on my local VM, the test manages to complete in about 4:40 minutes.

And the `passlib` test requires an `xfail` on linux with python 3.9, in which `crypt.crypt()` raises an `OSError` when libc does not support a particular encryption scheme (see [commit](https://github.com/python/cpython/commit/0d3fe8ae4961bf551e7d5e42559e2ede1a08fd7c) referenced from [here](https://bugs.python.org/issue39289)). The error seems to be raised on `passlib` import, so there's nothing we can really do about it in the test.

bwoodsend:
>  I wasn't even aware that Windows CI was running 32-bit version, though... So I've switched it over to 64-bit.

Wow. Me neither.

> The dash test definitely needs timeout larger than former 5 minutes, as on my local VM, the test manages to complete in about 4:40 minutes.

And are whatever enormousness files it's collecting supposed to be there? i.e. We shouldn't be adding `excludes` statements to the `dask` hook to try and tame it?

rokm:
> > The dash test definitely needs timeout larger than former 5 minutes, as on my local VM, the test manages to complete in about 4:40 minutes.
> 
> And are whatever enormousness files it's collecting supposed to be there? i.e. We shouldn't be adding `excludes` statements to the `dask` hook to try and tame it?

Hmm, we probably should be adding `excludeimports` *somewhere*...

This looks like one of those builds that collect a lot of stuff in a full-blown CI environment (matplotlib, IPython, core parts of PyQt5, core parts of PySide2, etc).

bwoodsend:
Ahh yes, the familiar `matplotlib` -> `matplotlib.backends` -> *every GUI/web-UI library in the known universe* trail...

Can you think of anything wrong with changing the matplotlib hook to just pick the first available backend? We could rank them in order of size so you get the smallest by default? Granted, it would give sub optimal behaviour if say you are already including PyQt5 in an application but PyInstaller gives you tkinter too but that's still better than the current behaviour of *include everything*.

rokm:
I suspect there are (at least) two chains involved in this particular case, and they both start with `plotly` (which is imported from `dash.dash`):

`dash.dash ` -> `plotly` ->  `plotly.utils` -> `_plotly_utils.data_utils` -> `PIL` -> `PIL.Image` -> `PIL.ImageQt` -> {`PyQt5`, `PyQt6`, `PySide2`, `PySide6`}

`dash.dash ` -> `plotly` -> {`plotly.data`, `plotly.express._core`, `plotly.express._imshow`, `plotly.figure_factory._county_choropleth`, `plotly.figure_factory._hexbin_mapbox`} -> `pandas` -> `pandas.core.config_init` -> `pandas.plotting` -> `pandas.plotting._core` -> `pandas.plotting._matplotlib` -> `matplotlib.pyplot` -> `matplotlib` -> ...

Our [new hook for `plotly`](https://github.com/pyinstaller/pyinstaller-hooks-contrib/blob/1649c5e412e02f5ebee4c542093064c8ba1c96bb/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-plotly.py#L17) even explicitly adds `pandas` to `hiddenimports`, although the above chain seems to get established even without that (e.g., when I was testing with latest released version without the hook instead of `master` branch).

rokm:
> Can you think of anything wrong with changing the matplotlib hook to just pick the first available backend?

I think better strategy would be to collect only the backend returned by `matplotlib.rcParams['backend']`. Of course, that won't work if user explicitly changes backend in their code via `matplotlib.use()`, but in that case, we could tell them to also set `MPLBACKEND` in their shell before running PyInstaller. 

Or, we could tie collection to MPLBACKEND environment variable itself (collect only the backend from variable if set, everything if it is not).

Or, perhaps the best approach would be to have PYI_MPLBACKEND_COLLECTION variable, with following values:
* `auto`: collect backend returned by `matplotlib.rcParams['backend']`
* `all`: collect all backends (current behavior)
* `none`: collect no backends, and have user deal with them via hiddenimports themselves (e.g., if they want to collect multiple backends to be able to switch between them)
* we could also allow actual backend names (or even comma-separated list) in here, in addition to the above three options

There's very likely no universal (automatic) solution to this problem, but that would make if configurable enough for different use cases...

PR #112, opened 2021-04-21
Added a hook for SunPy

jeffreypaul15:
Added a standard hook for [SunPy](https://github.com/sunpy/sunpy).

### DIFF
--- a/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-sunpy.py
+++ b/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-sunpy.py
@@ -1,0 +1,19 @@
+# ------------------------------------------------------------------
+# Copyright (c) 2020 PyInstaller Development Team.
+#
+# This file is distributed under the terms of the GNU General Public
+# License (version 2.0 or later).
+#
+# The full license is available in LICENSE.GPL.txt, distributed with
+# this software.
+#
+# SPDX-License-Identifier: GPL-2.0-or-later
+# ------------------------------------------------------------------
+
+from PyInstaller.utils.hooks import collect_data_files, collect_submodules
+
+# Importing non-python files
+datas = collect_data_files('sunpy')
+
+# Include imports which PyInstaller doesn't detect
+hiddenimports = collect_submodules('sunpy')
--- a/news/112.new.rst
+++ b/news/112.new.rst
@@ -1,0 +1,1 @@
+Added a standard hook for SunPy.


src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-sunpy.py
```diff
@@ -0,0 +1,19 @@
...
+# SPDX-License-Identifier: GPL-2.0-or-later
+# ------------------------------------------------------------------
+
+from PyInstaller.utils.hooks import collect_data_files, collect_submodules
+
+# Importing non-python files
+datas = collect_data_files('sunpy')
+
+# Include imports which PyInstaller doesn't detect
+hiddenimports = collect_submodules('sunpy')
```
bwoodsend, line 19: Given that sunpy a) has a lot of optional dependencies and b) has its tests inside the package, I wouldn't recommend doing this. With it, every single dependency, optional dependency, test dependency and dev tool will get sucked in. If you maintain a minimal virtual environment then it doesn't really matter but for anyone running in say full-blown conda then this hook would be a spectacular dependency bomb.

jeffreypaul15: Should I just import the required dependencies and not the test ones then?

rokm: What submodules does PyInstaller fail to detect?

jeffreypaul15: Sunpt import other sub-modules that are all part of sunpy, in a way that is not always auto-discovered by pyinstaller.

rokm: Let me rephrase the question then: are there many such missed submodules? 

PyInstaller usually fails to pick modules that are not directly imported (e.g., some indirect way of importing is done) or if there's cythonized modules in the mix. So it helps to know why exactly they aren't picked up automatically, in order to decide whether to go with `collect_submodules` or manualy include (and possibly also exclude) list.

bwoodsend: I see `sunpy` isn't pure Python. PyInstaller can't detect imports from inside Python extension modules. I also see one usage of `importlib` which is also invisible to PyInstaller. `collect_submodules()` does have a **filter** option. Excluding anything that looks like a test would get you some of the way there:
```python
collect_submodules("sunpy", filter=lambda x: "tests" not in x.split("."))
```

Another thing to consider is how lazy are `sunpy`'s `__init__.py`s when it comes to loading submodules? Does `import sunpy` implicitly load most if not all of `sunpy`'s submodules? Or is it *lazy* like `import scipy`, which loads almost nothing, instead expecting you to use `from scipy.category.algoritmn import feature_you_want` to load only the bit you're interested in and no more? If `sunpy` is lazy then PyInstaller would recognise this and do a nice job of bundling only the parts and dependencies you need. If you use `collect_submodules()` then this is lost. You get everything - whether you use it or not.

bwoodsend: As @rokm suggests, with a project that size and so many optional dependencies, I think you'd be better adding individual hidden imports as you need them. In theory, this should just be a case of scanning your C source (or pyx if your using Cython) for import statements and adding them.

jeffreypaul15: Thanks for the input @bwoodsend I shall leave this decision to the SunPy devs.

jeffreypaul15: @bwoodsend 
in one of the __init__.py files, there exists a : `rootdir = os.path.join(os.path.dirname(sunpy.__file__), "data", "test")`
with `rootdir` being part of the public objects in `__all__`.
PyInstaller fails to read this and I get an `AttributeError: module 'sunpy.data.test' has no attribute 'rootdir'`
even after adding sunpy.data as a hidden import.
This, and `get_distribution('sunpy').requires(['all'])` returns an error stating `Module 'sunpy' has no 'all'`, I've even added ` copy_metadata("sunpy")` like you've mentioned, this fixes the package metadata issue but this new one seems to be a problem.

bwoodsend: Never seen that one before. It appears to be a side effect of `copy_metadata()` (for reasons I don't know) renaming the `dist-info` folder to `egg-info`.
```python
>>> from PyInstaller.utils.hooks import copy_metadata
>>> copy_metadata("sunpy")
[('/e/notebooks/pyi-issues/112/env/lib/python3.9/site-packages/sunpy-2.1.4.dist-info', 'sunpy-2.1.4-py3.9.egg-info')]
```
If I rename it back to `sunpy-2.1.4.dist-info` then the problem goes away. I'll need to look into why on earth PyInstaller does this...

bwoodsend: Looks like pyinstaller/pyinstaller#1888 caused the change and this breakage is known and unresolved (pyinstaller/pyinstaller#3033). Yipeee...

jeffreypaul15: Yeah I came across that while looking for a fix, thanks for the help though! :)

PR #115, opened 2021-05-10
hooks: IPython: remove use of compat.modname_tkinter

rokm:
Use 'tkinter' instead.


rokm:
Related to the upcoming `tkinter`/`Tcl`/`Tk` cleanup on the `pyinstaller` side.

bwoodsend:
Any reason why you're only removing it there? I see it in other hooks and even in the docs.

rokm:
> Any reason why you're only removing it there? I see it in other hooks and even in the docs.

This was the only instance I could find in this repository.

PR #116, opened 2021-05-14
Update hook for pypylon to include data files

Unmoon:
#114 missed the data files required for discovering cameras.

Tested to work on a new Windows installation.

### DIFF
--- a/news/116.update.rst
+++ b/news/116.update.rst
@@ -1,0 +1,1 @@
+Update hook for PyPylon to include data files.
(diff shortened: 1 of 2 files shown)


bwoodsend:
We could really do with a test for this. Could you pick a hello world pypylon example which breaks if it doesn't have the hook?

Unmoon:
I think it's not possible to test this without having a physical Basler camera and a PC that has never connected to a Basler camera previously. I tried to test this on my PC, and after discovering a camera once, deleting the `ProducerCXP.cti` did not affect the discovery.

bwoodsend:
Fair enough.

PR #117, opened 2021-05-17
Add hook-pytds.py

allaboutmikey:
Simple copy_metadata to make python-tds work in packed form.


bwoodsend (changes requested):
Thanks @allaboutmikey. This needs a [news entry](https://github.com/pyinstaller/pyinstaller-hooks-contrib/blob/master/news/README.txt). And that hook needs the copyright header - just copy/paste it out of one of the other files.

PR #120, opened 2021-05-24
hooks: add hook for flirpy

diepholz:
Added a hook to pull in the necessary data files to use flirpy on a project.

### DIFF
--- a/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-flirpy.py
+++ b/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-flirpy.py
@@ -1,5 +1,5 @@
 # ------------------------------------------------------------------
-# Copyright (c) 2020 PyInstaller Development Team.
+# Copyright (c) 2021 PyInstaller Development Team.
 #
 # This file is distributed under the terms of the GNU General Public
 # License (version 2.0 or later).
--- a/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-flirpy.py
+++ b/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-flirpy.py
@@ -1,0 +1,19 @@
+# ------------------------------------------------------------------
+# Copyright (c) 2020 PyInstaller Development Team.
+#
+# This file is distributed under the terms of the GNU General Public
+# License (version 2.0 or later).
+#
+# The full license is available in LICENSE.GPL.txt, distributed with
+# this software.
+#
+# SPDX-License-Identifier: GPL-2.0-or-later
+# ------------------------------------------------------------------
+"""
+Hook for flirpy, a library to interact with FLIR thermal imaging cameras and images.
+https://github.com/LJMUAstroEcology/flirpy
+"""
+
+from PyInstaller.utils.hooks import collect_data_files
+
+datas = collect_data_files('flirpy')
--- a/src/_pyinstaller_hooks_contrib/tests/test_libraries.py
+++ b/src/_pyinstaller_hooks_contrib/tests/test_libraries.py
@@ -561,3 +561,10 @@ def test_blspy(pyi_builder):
     pyi_builder.test_source("""
         import blspy
         """)
+
+
+@importorskip('flirpy')
+def test_flirpy(pyi_builder):
+    pyi_builder.test_source("""
+        import flirpy
+        """)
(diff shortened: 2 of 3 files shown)


bwoodsend:
Could you try launching our manual CI?

* Go [here](https://github.com/diepholz/pyinstaller-hooks-contrib/actions/workflows/oneshot-test.yml)
* Click Run Workflow
* Set
  - the packages field to just `flirpy` (no quotes)
  - OS to `ubuntu, windows, macos`
  - Python versions to `3.6, 3.9`
* Then paste the url of the new job here.

diepholz:
sure thing - url of the new job below:

[](https://github.com/diepholz/pyinstaller-hooks-contrib/actions/runs/872014631)

src/_pyinstaller_hooks_contrib/tests/test_libraries.py
```diff
@@ -561,3 +561,10 @@ def test_blspy(pyi_builder):
     pyi_builder.test_source("""
         import blspy
         """)
+
+
+@importorskip('flirpy')
+def test_flirpy(pyi_builder):
+    pyi_builder.test_source("""
+        import flirpy
+        """)
```
rokm, line 570: `flirpy` has [empty top-level package](https://github.com/LJMUAstroecology/flirpy/blob/master/flirpy/__init__.py), so I don't think we're really testing anything here (i.e., the test passes even without the hook and without `flirpy/bin` being collected).

diepholz: okay, the usage in my actual application is:

`from flirpy.camera.lepton import Lepton
`

My thought was this was too specific, so tried to make the test case more general.  Open to suggestions on how you'd approach this.

bwoodsend: Specific is fine but maybe do several separate tests - one for each camera type? I'm worried that you may need a physical camera to run enough of `flirpy` to force it to try (and fail if it doesn't have the hook) to load these files in `flirpy/bin`.

rokm: I imagine you need the actual camera to test to the extent when the lack of collected `bin` folder becomes a problem (i.e., `from flirpy.camera.lepton import Lepton` doesn't trigger an error, either). As such, I think dropping the test altogether might be the best idea here.

bwoodsend: Given that `flirpy/bin/find_camera.exe` is loaded [here](https://github.com/LJMUAstroecology/flirpy/blob/a094fdeb09c6f775fd82e7f0287b461ab792382d/flirpy/camera/lepton.py#L33) perhaps:
```python
from flirpy.camera.lepton import Lepton

print(Lepton.find_video_device())
```
would be enough?

rokm: Indeed, that would do the trick. It returns `None` as opposed to raising exception when no camera is found, and `[WinError 2] The system cannot find the file specified` if `bin` is missing.

diepholz: Thanks folks - changes are pushed.

bwoodsend:
Cool. Rerun the CI then if that passes I'll hit the merge button.

src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-flirpy.py
```diff
@@ -0,0 +1,19 @@
+# ------------------------------------------------------------------
+# Copyright (c) 2020 PyInstaller Development Team.
```
rokm, line 2: Ah, missed this one earlier - should be 2021.

diepholz: I'll get that updated too.  Does the Copyright in test_libraries.py also need to be bumped to 2021?

rokm: > Does the Copyright in test_libraries.py also need to be bumped to 2021?

Actually, it looks like all files in this repository will need a copyright year update - it was done in the main repository, but apparently not here yet...

So for this PR, I'd just fix it in the new hook file, while the rest will be addressed separately.

rokm (approved):

bwoodsend:
I see [you're working on the CI run](https://github.com/diepholz/pyinstaller-hooks-contrib/runs/2667985238?check_suite_focus=true). If you cancel [those useless automatic runs](https://github.com/diepholz/pyinstaller-hooks-contrib/actions/workflows/macos-test.yml) then those macos jobs should get off the queue and start running.

bwoodsend:
[Happy days](https://github.com/diepholz/pyinstaller-hooks-contrib/actions/runs/872014631)...

PR #122, opened 2021-05-26
Add hook for humanize to collect package metadata

nicoddemus:
Taken from https://github.com/jmoiron/humanize/issues/105#issuecomment-594826403, credits go to @2press. 👍

### DIFF
--- a/news/122.new.rst
+++ b/news/122.new.rst
@@ -1,0 +1,1 @@
+Add hook for `humanize <https://pypi.org/project/humanize>`__ to include required metadata.
--- a/requirements-test-libraries.txt
+++ b/requirements-test-libraries.txt
@@ -5,6 +5,7 @@ boto3==1.12.33
 botocore==1.15.33
 dash==1.19.0
 dash-bootstrap-components==0.12.0
+humanize==3.5.0
 iminuit==2.4.0
 markdown==3.2.1
 openpyxl==3.0.3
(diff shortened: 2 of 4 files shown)


bwoodsend (approved):
Was in the midst of saying you've got a linter to appease but you beat me to it!

As a side note - as of pyinstaller/pyinstaller#5830 (i.e. the next release of PyInstaller) metadata collection should be automatic so there shouldn't be so many of these *include the metadata hooks* in future...

nicoddemus:
> As a side note - as of pyinstaller/pyinstaller#5830 (i.e. the next release of PyInstaller) metadata collection should be automatic so there shouldn't be so many of these include the metadata hooks in future...

Great news, thanks for the info. 👍

bwoodsend:
> Do you know if that will also work for setuptools entry points? This would make pytest plugins (for example) work out of the box, which would be neat.

No, I did consider doing entry points too but pussied out because I was worried that it would become a clutter magnet. Libraries such as NumPy for example reference `pytest` and therefore `import numpy` would imply *include all pytest's plugins*. We can batten down the hatches a bit by manually labeling NumPy as not dependent on pytest but it doesn't work too well.

There is a relatively new hook utility [collect_entry_point()](https://pyinstaller.readthedocs.io/en/stable/hooks.html#PyInstaller.utils.hooks.collect_entry_point) which I added to allow users to *opt in* to collecting plugins. Funnily enough, streamlining pytest plugin collection was what I had in mind when I wrote it.

nicoddemus:
> There is a relatively new hook utility collect_entry_point() which I added to allow users to opt in to collecting plugins. Funnily enough, streamlining pytest plugin collection was exactly what I wrote it for.

Great info, thanks. I will have to try that to replace the workarounds we had to do at work in order to freeze pytest + plugins. Awesome work! 👍

bwoodsend:
I see pyinstaller/pyinstaller#5830 is to blame is to blame for [that test failure](https://travis-ci.org/github/pyinstaller/pyinstaller-hooks-contrib/jobs/772514679). I better go mop that up before someone accidentally releases it...

PR #123, opened 2021-05-30
Create hook for Azurerm which is using pkg_resources internally

Hilali4886:
Azurerm is a lite api to microsoft azure.
Azurerm is using pkg_resources internally which is not supported by py-installer.
This hook will collect the module metadata.
Tested with Azurerm 0.10.0

### DIFF
--- a/news/123.new.rst
+++ b/news/123.new.rst
@@ -1,0 +1,1 @@
+Add a hook for ``Azurerm`` which is using pkg_resources internally.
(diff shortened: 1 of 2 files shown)


bwoodsend (changes requested):
We need all Python source files to have the copyright header. Otherwise anyone copying those files (including us) is technically infringing copyright and is liable to be sued. Copy/paste it from one of the other files.

And this needs a [news entry](https://pyinstaller.readthedocs.io/en/stable/development/changelog-entries.html).

bwoodsend:
Still need the news entry...

bwoodsend:
Thank you!

PR #125, opened 2021-06-12
Add office365 hook

colin-fsa:


src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-office365.py
```diff
@@ -0,0 +1,19 @@
...
+# ------------------------------------------------------------------
+
+"""
+Office365-REST-Python-Client contains xml templates that are needed by some methods
+This hook ensures that all of the data used by the package is bundled
+"""
+
+from PyInstaller.utils.hooks import collect_all
+
+datas, binaries, hiddenimports = collect_all("office365")
```
bwoodsend, line 19: `collect_all()` is extremely wasteful and should really only be used as a last resort. You only mention that it contains xml files so why not just use `collect_data_files()`?

CHANGELOG.rst
```diff
@@ -16,6 +16,8 @@ New hooks
   <https://github.com/pyinstaller/pyinstaller-hooks-contrib/issues/83>`_)
 * Added a hook for the 'pygraphviz' library (`#86
   <https://github.com/pyinstaller/pyinstaller-hooks-contrib/issues/86>`_)
+* Add a hook for ``Office365-REST-Python-Client`` which uses data files in
+  some methods
```
bwoodsend, line 20: This isn't how our changelog works. Editing the changelog directly invariably leads to merge conflicts when multiple PRs add to the same file in the same place. Instead write a [news entry](https://github.com/pyinstaller/pyinstaller-hooks-contrib/blob/master/news/README.txt).

bwoodsend (changes requested):
I'd also really like to see a test or two. That way we can send it to CI and test it on all platforms automatically.

CHANGELOG.rst
colin-fsa: Apologies for that, I had gotten confused by the instructions in the developer guide. Do you think we could rework that a bit to clarify expectations? I'd be happy to take a stab at revising it, just not sure how to do so.

src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-office365.py
colin-fsa: That works perfectly, thank you!

colin-fsa:
> I'd also really like to see a test or two. That way we can send it to CI and test it on all platforms automatically.

Sure thing! It looks like your test suite works by running code that will throw an exception without the hook so I went with that methodology. I found a way without creating a real request to reproduce the failure, the lines in the test print some stuff to stdout if the hook is present (success) and throw exceptions if the hook is absent (failure).

CHANGELOG.rst
bwoodsend: Hmm, I've got a half written branch collecting dust somewhere which would do it if I ever finish it. It's also going to explain how to write tests (which I see you've had to guess at - although you guessed right), how to trigger our custom CI/CD, and the copyright header (which most people forget).

src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-office365.py
```diff
@@ -0,0 +1,19 @@
...
+# ------------------------------------------------------------------
+
+"""
+Office365-REST-Python-Client contains xml templates that are needed by some methods
+This hook ensures that all of the data used by the package is bundled
+"""
+
+from PyInstaller.utils.hooks import collect_data_files
+
+datas = collect_data_files("office365")
```
bwoodsend, line 19: Files need to end with a blank line or the linter cries.

bwoodsend (changes requested):
Nice. Next you've got a [linter to appease](https://github.com/pyinstaller/pyinstaller-hooks-contrib/pull/125/checks?check_run_id=2811075516#step:6:53). I suggest that you run it locally otherwise Github's new *approve CI run* feature will drive us both nuts.
```
pip install flake8
git diff -U0 origin/master -- | flake8 --diff -
```
Depending on how your remotes are set up, you might need to replace `origin/master` with the last commit ID that you didn't write yourself.

bwoodsend (approved):
[Tests pass](https://github.com/pyinstaller/pyinstaller-hooks-contrib/actions/runs/932027863) too...

bwoodsend:
Nice work @colin-fsa

colin-fsa:
Thanks @bwoodsend! Appreciate all of your assistance with learning the ropes.

PR #130, opened 2021-06-17
Add a guide to contributing hooks.

bwoodsend:
As the title suggests, include a guide for contributing hooks in the README. 

This ended up quite a bit longer than I was expecting - by which I mean that I think it's too long. I'd forgotten just how many hoops we have to jump through.

@colin-fsa As you're the most recent person to ask me the *are any of these steps written down somewhere?* question, I'd appreciate your review on this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,5 @@
 # `pyinstaller-hooks-contrib`: The PyInstaller community hooks repository
 
-
-
 What happens when (your?) package doesn't work with PyInstaller? Say you have data files that you need at runtime?
 PyInstaller doesn't bundle those. Your package requires others which PyInstaller can't see? How do you fix that?
 
@@ -12,6 +10,7 @@ such as the data files and (hidden) imports mentioned above.
 
 This repository is a collection of hooks for many packages, and allows PyInstaller to work with these packages
 seamlessly.
+
 
 ## Installation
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -28,13 +27,297 @@ pip install -U pyinstaller-hooks-contrib
...
+* [ ] Your changes [satisfy the linter](#run-linter) (run `git diff -U0 master | flake8 --diff -`).
+
+
+### Submit the pull request
+
+Once you've done all the above, go ahead and create a pull request.
+If you're stuck doing any of the above steps, create a draft pull request and explain what's wrong - we'll sort you out...
+Feel free to copy/paste commit messages into the Github pull request title and description.
+If you have run CI/CD, please include a link to it in your description so that we can see that it works.
+If you've never done a pull request before, note that you can edit it simply by running `git push` again.
```
rokm, line 317: It might be worth mentioning that `git push` works if one adds new commits to the branch, whereas updating existing commits requires `git push --force`.

rokm (approved):
That's a nice and detailed writeup of all the required steps (and as you said, there *are* a lot of hoops to go through).

README.md
bwoodsend: Actually, one nasty git operation for noobs which I've forgotten about: Once you've got one hook merged, how to update your fork to match upstream so that you can do a second?
```
git checkout master
git pull git@github.com:pyinstaller/pyinstaller-hooks-contrib.git
```
does the job but you need to understand how remotes work to know that.

README.md
```diff
@@ -28,13 +27,297 @@ pip install -U pyinstaller-hooks-contrib
...
+**Please provide one hook per pull request! If you have more than one then submit them in separate pull requests.**
+
+
+### Setup
+
+[Fork this repo](https://github.com/pyinstaller/pyinstaller-hooks-contrib/fork) if you haven't already done so.
+Clone and `cd` inside your fork by running the following (replacing `bob-the-barnacle` with your github username):
+
+```
+git clone git@github.com:bob-the-barnacle/pyinstaller-hoooks-contrib.git
```
bwoodsend, line 48: Do you think it would be more sensible to show an https url? I think someone who wants to use ssh is more likely to know to switch the url format than the other way around.

README.md
bwoodsend: See as we always squash merge (unless someone who knows git well enough to mop up their own history submits multiple hooks in one PR) I was rather hoping that there would never be any need to use `commit --amend` or rebase and therefore never any need to force push. That said, I'm sure someone will somehow prove me wrong the hard way before long.

bwoodsend:
Grr, mental note to never reply to a review comment from the `files changed` tab. Now my replies are all over the place.

README.md
rokm: > > It might be worth mentioning that git push works if one adds new commits to the branch, whereas updating existing commits requires git push --force.
> 
> See as we always squash merge (unless someone who knows git well enough to mop up their own history submits multiple hooks in one PR) I was rather hoping that there would never be any need to use `commit --amend` or rebase and therefore never any need to force push. That said, I'm sure someone will somehow prove me wrong the hard way before long.

OK, I guess amending commits and force-pushing do indeed fall out of scope here.

rokm: > Actually, one nasty git operation for noobs which I've forgotten about: Once you've got one hook merged, how to update your fork to match upstream so that you can do a second?
> 
> ```
> git checkout master
> git pull git@github.com:pyinstaller/pyinstaller-hooks-contrib.git
> ```
> 
> does the job but you need to understand how remotes work to know that.

I think for complete newbies, the GitHub interface also offers a way to sync one's fork with upstream. 

But at least mentioning the CLI way (and perhaps giving a reference to more comprehensive tutorial about remotes, pull vs. pull --rebase, etc.) isn't a bad idea, either.

README.md
```diff
@@ -28,13 +27,297 @@ pip install -U pyinstaller-hooks-contrib
...
+    For bonus points, track down why so many submodules are hidden. Typical causes are:
+    *   Lazily loaded submodules (`importlib.importmodule()` inside a module `__getattr__()`).
+    *   Dynamically loaded *backends*.
+    *   Usage of `Cython` or Python extension modules containing `import` statements.
+*   Use of [`collect_all()`](https://pyinstaller.readthedocs.io/en/latest/hooks.html#PyInstaller.utils.hooks.collect_all).
+    This function is extremely inefficient and is really only for emergencies.
+    Check that you really do need to collect all of submodules, data files, binaries, metadata and dependencies.
+    If you do then add a comment to say so (and if you know it - why).
+    Do not simply use `collect_all()` just to *future proof* the hook.
+*   Any complicated `os.path` arithmetic.
```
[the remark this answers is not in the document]
rokm: I'll pass judgment on that one, because I've seen enough `os.path` arithmetic around here that I know exactly what it means...

README.md
rokm: It might be; if nothing else, GitHub also offers https url by default. Plus, if someone had set keys for ssh access, they will probably know to switch from https url to ssh one.

bwoodsend:
@rokm What do you want to do about the `requirements-libraries.txt`? I've generally (albeit inconsistently) been choosing not to add new libraries in because I didn't like our old per pull request CI just getting longer and longer but now that it's scheduled weekly I reckon we should be adding them in. Possibly even let Pyup loose on it?

README.md
bwoodsend: I've pointed out the **Fetch upstream** button Github now has. I thought adding commands in there would make it look more complicated than it is.

rokm:
> @rokm What do you want to do about the `requirements-libraries.txt`? I've generally (albeit inconsistently) been choosing not to add new libraries in because I didn't like our old per pull request CI just getting longer and longer but now that it's scheduled weekly I reckon we should be adding them in. Possibly even let Pyup loose on it?

I think the only value of having libraries in there is if we plan to use our weekly CI runs as early-warning breakage detector for new versions of those libraries. But for that to work, we'll need Pyup to keep updating the pinned versions. (Although Pyup's PRs could also trigger CI, so we don't blindly repin to incompatible versions?).

bwoodsend:
> I think the only value of having libraries in there is if we plan to use our weekly CI runs as early-warning breakage detector for new versions of those libraries. But for that to work, we'll need Pyup to keep updating the pinned versions. (Although Pyup's PRs could also trigger CI, so we don't blindly repin to incompatible versions?).

That is roughly what I have in mind. I can set our old per pull request CI/CD workflows to only run on pull requests from the pyup bot which should more or less achieve what we're after. My only issue with Pyup is that it doesn't respect our supported Python versions so there can be a bit of hairiness as packages fail to install on recently dropped Pythons but it's a minor issue.

README.md
```diff
@@ -28,13 +27,298 @@ pip install -U pyinstaller-hooks-contrib
...
+This is for the same reason we filter tests when running locally -
+the full test suite runs like a dead snail on ketamine.
+
+First push the changes you've made so far.
+
+```commandline
+git push --set-upstream origin hook-for-foo
+```
+
+Replace *billy-the-buffalo* with your Github username in the following url then open it.
```
colin-fsa, line 222: Needs to be updated, there's no billy-the-buffalo on the line above.

README.md
```diff
@@ -28,13 +27,298 @@ pip install -U pyinstaller-hooks-contrib
...
+It should take you to the `oneshot-test` actions workflow on your fork.
+You may be asked if you want to enable actions on your fork - say yes.
+```
+https://github.com/billy-the-buffalo/pyinstaller-hooks-contrib/actions/workflows/oneshot-test.yml
+```
+
+Find the **Run workflow** button and click on it.
```
colin-fsa, line 229: You need to click on 'Oneshot test' on the left to have the run button appear. I'd clarify not to click the 'New Workflow' button since that was my best guess at the next step after I couldn't find the run button.

README.md
```diff
@@ -28,13 +27,298 @@ pip install -U pyinstaller-hooks-contrib
...
+A dialog should appear containing one drop-down menu and 5 line-edit fields.
+This dialog is where you specify what to test and which platforms and Python versions to test on.
+Its fields are as follows:
+1.  A branch to run from. Set this to the branch which you are using (e.g. ``hook-for-foo``),
```
colin-fsa, line 233: Adding this to look at later - I believe the pyinstaller dev guide specifies to work on your fork's master branch, presumably if you're suggesting other branch names it's because this guide recommends branching somewhere.

README.md
```diff
@@ -28,13 +27,298 @@ pip install -U pyinstaller-hooks-contrib
...
+2.  Which package(s) to install and their version(s).
```
colin-fsa, line 234: Maybe note to refer to your own commit's entries in requirements-test-libraries.txt? I just did that and noticed I had missed adding libraries.

README.md
```diff
@@ -28,13 +27,298 @@ pip install -U pyinstaller-hooks-contrib
...
+    * Set to `foo` to test the latest version of `foo`,
+    * Set to `foo==1.2, foo==2.3` (note the comma) to test two different versions of `foo` in separate jobs,
+    * Set to `foo bar` (note the lack of a comma) to test `foo` and `bar` in the same job,
+3.  Which OS or OSs to run on
+    * Set to `ubuntu` to test only `ubuntu`,
+    * Set to `ubuntu, macos, windows` (order is unimportant) to test all three OSs.
+4.  Which Python version(s) to run on
+    * Set to `3.9` to test only Python 3.9,
+    * Set to `3.6, 3.7, 3.8, 3.9` to test all currently supported version of Python.
+5.  The final two options can generally be left alone.
```
colin-fsa, line 245: Since this is a step-by-step guide, it would be helpful to say before the step that you don't need to follow it. I was following along and got confused about how I was supposed to make the decisions for steps 3 and 4.

README.md
bwoodsend: The url (having the `workflows/oneshot-test.yml` suffix) should've taken care of that. Did you have to click the *enable actions on this fork* page and that didn't forward you directly to the workflow?

README.md
bwoodsend: > I believe the pyinstaller dev guide specifies to work on your fork's master branch

If it does then it needs changing. Certainly I think the current maintainers all encourage creating branches. And yes, there's a recommendation to do so [at the top](https://github.com/pyinstaller/pyinstaller-hooks-contrib/pull/130/commits/5ac583f7e3f48ef142655ffd6c325a8020c81493#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R53-R58).

README.md
bwoodsend: Sure. First I need to add something about adding to `requirements-test-libraries.txt` to begin with...

README.md
bwoodsend: What do you mean by you don't need to follow it? Do you just mean that the defaults will work or something else?

README.md
```diff
@@ -28,13 +27,298 @@ pip install -U pyinstaller-hooks-contrib
...
+Find the **Run workflow** button and click on it.
+A dialog should appear containing one drop-down menu and 5 line-edit fields.
+This dialog is where you specify what to test and which platforms and Python versions to test on.
+Its fields are as follows:
+1.  A branch to run from. Set this to the branch which you are using (e.g. ``hook-for-foo``),
+2.  Which package(s) to install and their version(s).
+    Which packages to test are inferred from which packages are installed.
+    * Set to `foo` to test the latest version of `foo`,
+    * Set to `foo==1.2, foo==2.3` (note the comma) to test two different versions of `foo` in separate jobs,
+    * Set to `foo bar` (note the lack of a comma) to test `foo` and `bar` in the same job,
```
colin-fsa, line 238: Is this right? I did that and got `ERROR: Invalid requirement: 'spacy srsly thinc'`: https://github.com/colin-fsa/pyinstaller-hooks-contrib/runs/2867341158?check_suite_focus=true

README.md
colin-fsa: Sorry, you're right. Hit send too quickly.

README.md
colin-fsa: I think I misread this the first time. It's worded fine.

README.md
colin-fsa: Yes, that's why

README.md
bwoodsend: Looks like you wrapped used `"spacy srsly thinc"` instead of just `spacy srsly thinc`?

colin-fsa: That seemed like the obvious thing to do since the default for that input is `"numpy==1.19", "numpy<1.18"`

README.md
```diff
@@ -28,13 +27,298 @@ pip install -U pyinstaller-hooks-contrib
...
+* [ ] Your changes [satisfy the linter](#run-linter) (run `git diff -U0 master | flake8 --diff -`).
+
+
+### Submit the pull request
+
+Once you've done all the above, go ahead and create a pull request.
+If you're stuck doing any of the above steps, create a draft pull request and explain what's wrong - we'll sort you out...
+Feel free to copy/paste commit messages into the Github pull request title and description.
+If you have run CI/CD, please include a link to it in your description so that we can see that it works.
+If you've never done a pull request before, note that you can edit it simply by running `git push` again.
```
colin-fsa, line 318: Not to rehash the whole discussion you already had with @rokm, but I think you should at least mention rebasing and force pushing. I had never used either of those things before working on this project and the majority of my time has been spent trying to clean my commits up, so I suspect others would also appreciate some advice or at least training resources.

README.md
```diff
@@ -28,13 +27,298 @@ pip install -U pyinstaller-hooks-contrib
...
+```
+
+Note that on macOS and Linux, `pip` may by called `pip3`.
+If you normally use `pip3` and `python3` then use `pip3` here too.
+You may skip the 2<sup>nd</sup> line if you have no intention of providing tests (but please do provide tests!).
+
+
+### Add the hook
+
+Hooks live in the [src/_pyinstaller_hooks_contrib/hooks/stdhooks/](../master/src/_pyinstaller_hooks_contrib/hooks/stdhooks/) directory.
```
colin-fsa, line 79: Just wanted to confirm it was a purposeful decision to not talk about runtime hooks here

README.md
```diff
@@ -28,13 +27,298 @@ pip install -U pyinstaller-hooks-contrib
...
+Simply copy your hook into there.
```
colin-fsa, line 80: Would it be worth giving guidance on how to get started creating a hook if you have absolutely no idea? Like:
```
A common technique for creating a hook is to get the hook working as quickly as possible, then narrow down the minimal logic to keep it working:

- [ ] Write a test, ie. test_foo(). If you're not sure how to write a test try _import foo_, or a minimal reproduceable example of the code that caused you to determine the hook is needed
- [ ] Run the test and ensure it fails: _pytest -k test_foo_ 
- [ ] Create a new hook file and populate it with:
     _from PyInstaller.utils.hooks import collect_all_
    _datas, binaries, hiddenimports = collect_all("foo")_
- [ ] Rerun your test
    If it fails: Additional work will be needed. If you require assistance please feel free to submit a draft and ask for help.
- [ ] If the test passes, replace your existing hook with:
    _datas = copy_metadata('foo')
    binaries = collect_dynamic_libs('foo')
    hiddenimports = collect_submodules('foo')_
- [ ] If the tests still passes, determine if any of these lines can be removed to create a minimal hook
```

README.md
```diff
@@ -28,13 +27,298 @@ pip install -U pyinstaller-hooks-contrib
...
 
 
 ## I want to help!
 
-Please start by providing pull requests and helping solve issues.
+If you've got a hook you want to share then great!
+The rest of this page will walk you through the process of contributing a hook.
+If you've been here before then you may want to skip to the [summary checklist](#summary)
+
+**Please provide one hook per pull request! If you have more than one then submit them in separate pull requests.**
```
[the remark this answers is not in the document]
bwoodsend: See normally I'd still rather they come as separate PRs which just cross reference each other. It's mostly just so that we can squash merge each one and therefore is doesn't matter if the PR author makes a mess of git history.

README.md
bwoodsend: Yeah I see where you're coming from. It does appear to imply that each job should be quoted. The quotes are only needed to stop bash from interpreting `<` as a redirection. Other than that, they should never be needed. I'm thinking perhaps I could have it escape those characters itself so that there are no quotes involved at all...

README.md
bwoodsend: I'd quite like something like that on the [page about hooks in the docs](https://pyinstaller.readthedocs.io/en/latest/hooks.html) rather than all the drivel that's on that page currently. Then this readme can just reference it.

README.md
```diff
@@ -28,13 +27,298 @@ pip install -U pyinstaller-hooks-contrib
...
+# Copyright (c) 2021 PyInstaller Development Team.
+#
+# This file is distributed under the terms of the GNU General Public
+# License (version 2.0 or later).
+#
+# The full license is available in LICENSE.GPL.txt, distributed with
+# this software.
+#
+# SPDX-License-Identifier: GPL-2.0-or-later
+# ------------------------------------------------------------------
```
Legorooj, line 137: You appear to have duplicated the headers here - this one should be Apache, I believe.

README.md
bwoodsend: I like to pretend that runtime hooks don't exist most of the time. 423b95f7b9f8ca626773e5e22424419a4fcb65ea adds a line on runtime hooks.

README.md
bwoodsend: Fixed.

README.md
bwoodsend: Reckon 7f092f00acf50039f6b199a65103a142533658bc would have kept you right the first time?

README.md
```diff
@@ -28,13 +27,314 @@ pip install -U pyinstaller-hooks-contrib
...
+
+        # Your test here!
+        import foo
+
+        foo.something_fooey()
+
+    """)
+```
+
+If need be, you can specify version constraints using `@importorskip('foo', minversion="1.2.3")`
```
wmmc88, line 191: I think the `minversion` in the `importorskip` no longer works since https://github.com/pyinstaller/pyinstaller/pull/6023. Should it mention `@requires` instead?

bwoodsend: Yes, you're right. I've been meaning to address that for ages.

PR #134, opened 2021-06-24
Adds hook for sunpy

jeffreypaul15:
From the suggestions over at #112 
This hook only contains the files and imports which do not any tests.

### DIFF
--- a/news/134.new.rst
+++ b/news/134.new.rst
@@ -1,0 +1,1 @@
+Added a standard hook for SunPy.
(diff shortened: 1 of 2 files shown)


bwoodsend (changes requested):
OK, give it a [news entry](https://github.com/pyinstaller/pyinstaller-hooks-contrib/blob/master/news/README.txt) and maybe stick a comment in the hook to say *this is just a legacy hook - sunpy >= x.y.z ships its own hooks*.

Then we'll call it a day...

jeffreypaul15:
> OK, give it a [news entry](https://github.com/pyinstaller/pyinstaller-hooks-contrib/blob/master/news/README.txt) and maybe stick a comment in the hook to say _this is just a legacy hook - sunpy >= x.y.z ships its own hooks_.
> 
> Then we'll call it a day...

Sure, thanks for the help

PR #139, opened 2021-06-30
Msoffcrypto

colin-fsa:
I'm having trouble figuring out how to handle this one and was hoping I could get some advice. The offending line from the package seems to be `__version__ = pkg_resources.get_distribution("msoffcrypto-tool").version`. The version was a hardcoded string in the prior release and everything worked fine then, here's the commit that broke it: https://github.com/nolze/msoffcrypto-tool/commit/161b182a029ced7597c2c5240ed8348467ffeeb9#diff-589d1c7aac0cfc6ed2f443b43655fcaad7fee63a261631470e74e01de91d2e24

My draft hook doesn't help at all, I just left it in the draft PR to show that I had already tried the kitchen sink approach.

### DIFF
--- a/news/139.new.rst
+++ b/news/139.new.rst
@@ -1,0 +1,1 @@
+Add a hook for ``msoffcrypto`` to collect metadata.
--- a/requirements-test-libraries.txt
+++ b/requirements-test-libraries.txt
@@ -9,6 +9,7 @@ dash-bootstrap-components==0.12.0
 humanize==3.5.0
 iminuit==2.4.0
 markdown==3.2.1
+msoffcrypto-tool==4.12.0
 Office365-REST-Python-Client==2.3.4
 openpyxl==3.0.3
 passlib==1.7.2
--- a/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-msoffcrypto.py
+++ b/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-msoffcrypto.py
@@ -1,0 +1,18 @@
+# ------------------------------------------------------------------
+# Copyright (c) 2021 PyInstaller Development Team.
+#
+# This file is distributed under the terms of the Apache License 2.0
+#
+# The full license is available in LICENSE.APL.txt, distributed with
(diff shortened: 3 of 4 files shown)


rokm:
`datas = copy_metadata('msoffcrypto-tool')` should take care of that.

bwoodsend:
Use of `pkg_resources.get_distribution("msoffcrypto-tool")` translates as *collect package metadata for `msoffcrypto-tool`* so `datas = copy_metadata("msoffcrypto-tool")` should do it. But `collect_all("msoffcrypto-tool")` includes collecting metadata so it's odd that it isn't enough.

rokm:
> But `collect_all("msoffcrypto-tool")` includes collecting metadata so it's odd that it isn't enough.

That's because `collect_all` is done on "msoffcrypto", not "msoffcrypto-tool".

bwoodsend:
So it is...

colin-fsa:
Fascinating. I always seem to get into trouble with these packages that have different names for installation and importing. Thanks for the help! I'll clean up the PR, but more importantly I can get my actual application working now.

colin-fsa:
https://github.com/colin-fsa/pyinstaller-hooks-contrib/actions/runs/986750926
[Oneshot test](https://github.com/colin-fsa/pyinstaller-hooks-contrib/actions/workflows/oneshot-test.yml)

bwoodsend:
Yeah the line between distributions which you install and packages which you import is a sneaky one. And there are corners of PyInstaller which are blissfully unaware of the distinction. For example `collect_all()` uses:
* `collect_submodules()` `collect_data_files()` and `collect_dynamic_libs()` which assume the argument is an importable package
* `copy_metadata()` which assumes the argument is a distribution
* `requirements_for_package()` which assumes a package, tries to convert to distribution, gets the distribution dependencies then converts each of those back into packages, taking a ridiculously long time to do it

So no matter what you give it, `collect_all()` will never work completely if the package and distribution don't match.

bwoodsend:
Cheers Colin!

PR #276, opened 2021-07-19
hooks: scikit-image & scikit-learn: use distribution names for version checks

rokm:
Use distribution names (`scikit_image` and `scikit_learn`) for version checks via `is_module_satisfies()` instead of package names (`skimage` and `sklearn`). It is possible to install latest `scikit-learn` by installing 'sklearn', which installs `sklearn==0.0` that ends up messing up the version checks.

Make same adjustment in `scikit-image` hooks for the sake of consistency (although trying to install `skimage` shows an error     telling user to install `scikit-image` instead).

### DIFF
--- a/news/276.update.rst
+++ b/news/276.update.rst
@@ -1,0 +1,3 @@
+Update ``scikit-learn`` and ``scikit-image`` hooks to perform version checks 
+based on distribution name instead of package name, to prevent failures
+when ``sklearn`` dummy distribution is installed.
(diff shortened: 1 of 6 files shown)


rokm:
This is something that I stumbled into while trying to reproduce pyinstaller/pyinstaller#6035 and accidentally installed `scikit-learn` via `pip install sklearn`. Due to `sklearn==0.0` ending up installed (along with latest version of `scikit-learn`, the corresponding version checks failed and first caused the `test_sklearn` to be skipped, and afterwards fail with missing modules in a very similar way as reported in pyinstaller/pyinstaller#6035.

bwoodsend (approved):

PR #279, opened 2021-07-27
Add hook-mariadb.py

i3rinkmann:


src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-mariadb.py
```diff
@@ -0,0 +1,20 @@
+# ------------------------------------------------------------------
+# Copyright (c) 2020 PyInstaller Development Team.
```
rokm, line 2: Please use the current year here.

src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-mariadb.py
```diff
@@ -0,0 +1,20 @@
...
+# ------------------------------------------------------------------
+
+
+from PyInstaller.utils.hooks import get_pyextension_imports
+
+
+# The MariaDB uses a .pyd file and uses import within its __init__.py
+# I looked for similar hooks and found that pyodbc uses this function
+# for the import.
+hiddenimports = get_pyextension_imports('mariadb')
```
rokm, line 20: Add an empty line at the end of file.

rokm:
What OS and what version of `mariadb` are you having problems with (i.e., requires the hidden imports)? 

The simple import test seems to work out-of-box with 1.0.7 on both Windows and linux...

i3rinkmann:
I tried to create the exe on Windows 10 with python 3.7.8, mariadb 1.0.7 and pyinstalller 4.4

i use pipenv to manage my virtual env

It only crashed at runtime of the exe, not in the build process itself.

If it helps it crashed in the `__init__.py` from mariadb at line 10 where it imports `from ._mariadb` a whole bunch of modules. As far as i can tell those are all bundled in the `_mariadb.cp37-win_amd64.pyd`.

rokm:
> I tried to create the exe on Windows 10 with python 3.7.8, mariadb 1.0.7 and pyinstalller 4.4

Hmm, it looks like it depends on python version. It works with 3.8 and 3.9, but not with 3.7 nor 3.6. And it seems that `decimal` is the missing module, so just adding that to `hiddenimports` in the hook should suffice.

Using `get_pyextension_imports` is somewhat inconsistent because with recent `mariadb` versions, `mariadb` itself is a package rather than extension, but because the way the function works, it still gets the job done. It seems a bit heavy-handed, though, as we're seemingly missing only `decimal`. @bwoodsend, how flexible/future-proof do you think we need to be here?

bwoodsend:
I don't really mind. Once PyUp is going we shouldn't need to future proof stuff because we'll be notified almost immediately if a new hidden import appears. That said, I don't see any harm in using `get_pyextension_imports()` - other than its ugliness.

bwoodsend (changes requested):
Can you add `mariadb` and its version to the [requirements-test-libraries.txt](https://github.com/pyinstaller/pyinstaller-hooks-contrib/blob/master/requirements-test-libraries.txt).

i3rinkmann:
It seems only importing decimal in the hidden imports is sufficient.

is there a clean/consistent way to check for the python version so this import is only done on python 3.7 and older?
If such a check is used elsewhere in the project i could use that so it behaves the same within the project.

@bwoodsend will do

rokm:
> is there a clean/consistent way to check for the python version so this import is only done on python 3.7 and older?

You can use `is_pyXX` from [PyInstaller.compat](https://pyinstaller.readthedocs.io/en/latest/hooks.html?highlight=is_py#module-PyInstaller.compat):

```python
from PyInstaller.compat import is_py38

if not is_py38:
    hiddenimports = ['decimal']
```

bwoodsend (approved):
Looks good provided that [this workflow](https://github.com/pyinstaller/pyinstaller-hooks-contrib/actions/runs/1071634516) passes.

rokm:
> Looks good provided that [this workflow](https://github.com/pyinstaller/pyinstaller-hooks-contrib/actions/runs/1071634516) passes.
> 
> Edit: maybe not that one then. I see mariadb isn't just a simple pip install.

On Windows, you get binary wheel, but on other platforms the extension needs to be compiled, which requires `mariadb` development package.

bwoodsend:
In which case, it's probably not a good idea to have `mariadb` in the requirements file after all.

rokm (approved):

bwoodsend:
Or perhaps we should make it Windows only?

rokm:
> In which case, it's probably not a good idea to have `mariadb` in the requirements file after all.

Hmm, on linux, we could install `libmariadbclient-dev ̇`. But we'll need to exclude the package from requirements on macOS.

bwoodsend:
Let's go with that then. @i3rinkmann Can you add the following line [here](https://github.com/pyinstaller/pyinstaller-hooks-contrib/blob/master/.travis.yml#L25):
```
        - apt-get install -y libmariadbclient-dev
```
Then change that line in the requirements file to exclude macOS (add `; sys_platform != "darwin"`).

PR #280, opened 2021-07-30
Add hook-dash_uploader.py

eric15342335:
Add a hook for `dash-uploader` to collect data files

### DIFF
--- a/src/_pyinstaller_hooks_contrib/tests/test_libraries.py
+++ b/src/_pyinstaller_hooks_contrib/tests/test_libraries.py
@@ -634,3 +634,12 @@ def test_mariadb(pyi_builder):
     pyi_builder.test_source("""
         import mariadb
         """)
+
+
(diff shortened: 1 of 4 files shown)


src/_pyinstaller_hooks_contrib/tests/test_libraries.py
```diff
@@ -634,3 +634,12 @@ def test_mariadb(pyi_builder):
     pyi_builder.test_source("""
         import mariadb
         """)
+
+
+@importorskip('dash_uploader')
+def test_dash_uploader(pyi_builder):
+    pyi_builder.test_source("""
+        import dash
+        import dash_html_components
```
bwoodsend, line 643: What's the motivation behind these two lines?

eric15342335: That was a mistake. I will correct it now.

PR #282, opened 2021-08-01
Update hook-azurerm.py

Hilali4886:
Fix for pyinstaller4.4 need to use recursive=True


bwoodsend (changes requested):
Unfortunately, that **recursive** option is only available in PyInstaller 4.4 and later so it would have to be conditional to support both old and new versions of PyInstaller.
```python
from PyInstaller.utils.hooks import is_module_satisfies:

if is_module_satisfies("pyinstaller >= 4.4"):
    datas = copy_metadata("azurem", recursive=True)
else:
    datas = copy_metadata("azurem")
```

PR #286, opened 2021-08-05
hooks: add hook for pyqt5_plugins

eric15342335:
pkg_resources [line](https://github.com/altendky/pyqt-plugins/blob/main/src/pyqt_plugins/__init__.py#L11)
come from https://stackoverflow.com/questions/68665420

### DIFF
--- a/requirements-test-libraries.txt
+++ b/requirements-test-libraries.txt
@@ -26,6 +26,9 @@ pycparser==2.20
 pycryptodome==3.9.7
 pycryptodomex==3.9.7
 pyexcelerate==0.8.0
+PyQt5==5.15.4
+qt5-applications==5.15.2.2.2
+qt5-tools==5.15.2.1.2
 pylint==2.4.4
 pyusb==1.0.2
 pyzmq==22.0.3
--- a/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-pyqt5_plugins.py
+++ b/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-pyqt5_plugins.py
@@ -1,0 +1,17 @@
+# ------------------------------------------------------------------
+# Copyright (c) 2020 PyInstaller Development Team.
+#
+# This file is distributed under the terms of the GNU General Public
+# License (version 2.0 or later).
+#
+# The full license is available in LICENSE.GPL.txt, distributed with
+# this software.
+#
+# SPDX-License-Identifier: GPL-2.0-or-later
+# ------------------------------------------------------------------
+
+from PyInstaller.utils.hooks import copy_metadata, collect_submodules
+# Those modules uses pkg_resources.get_distribution(foo()) which are not detectable via bytecode scanning
+datas = [copy_metadata("pyqt5_plugins"), copy_metadata("qt5_tools"), copy_metadata("qt5_applications")]
+# Lazily loaded submodules ( importlib.import_module() )
+hiddenimports = ["qt5_tools", collect_submodules("pyqt5_plugins"), collect_submodules("qt5_applications")]
--- a/src/_pyinstaller_hooks_contrib/tests/test_libraries.py
+++ b/src/_pyinstaller_hooks_contrib/tests/test_libraries.py
@@ -657,3 +657,12 @@ def test_mnemonic(pyi_builder):
         import mnemonic
         mnemonic.Mnemonic("english")
         """)
+
+
+@importorskip('pyqt5_plugins')
+def test_pyqt5_plugins(pyi_builder):
+    pyi_builder.test_source("""
+        import PyQt5
+        import PyQt5.QtCore
+        import pyqt5_plugins
+        """)


rokm:
Huh... as if the whole Qt+pyinstaller situation was not messy enough by itself... 

---------
This looks like it will require several hooks that target the following simple test program:

```python
import pyqt5_plugins
``` 

hook-pyqt5_plugins.py:
 * copy_metadata for itself (due to already discovered [pkg_resources line](https://github.com/altendky/pyqt-plugins/blob/8138b740346f475181a94729832a6095e0ac1d24/src/pyqt_plugins/__init__.py#L11))
 * hiddenimports (due to [these](https://github.com/altendky/pyqt-plugins/blob/8138b740346f475181a94729832a6095e0ac1d24/src/pyqt_plugins/__init__.py#L30-L34)):
   * `PyQt5.Core`
   * `qt5_tools`
   * `pyqt5_plugins.utilities` (and perhaps also `pyqt5_plugins._version`; or just brute-force collect_submodules of `pyqt5_plugins`)

hooks-qt5_tools.py:
 * copy_metadata for itself (due to [pkg_resources line here](https://github.com/altendky/qt-tools/blob/e81ea193a28a060875bba18b0a2f23c5506e0822/src/qt_tools/__init__.py#L11), which is triggered by trying to import `qt_applications` via `import_it`)
 * hidden import of `qt5_applications` (due to import via `import_it`)
 * perhaps we'd need to collect some modules from the package (like `_version`) or just brute force it with `collect_submodules`

hooks-qt5_applications.py:
 * copy_metadata for itself (due to [pkg_resource line here](https://github.com/altendky/qt-applications/blob/2a585266f2c8ef944766d15c582b76b93753155d/src/qt_applications/__init__.py#L9), triggered by attempting to import `_application` from itself - makes one wonder why regular relative import is not good enough...)
* hidden import of `qt5_applications._application` (due to import via `import_it`)
* perhaps we'd need to collect some other modules from the package (like `_version`) or just brute force it with `collect_submodules`

--------

There exist `pyqt6_plugins`, `qt6_tools`, and `qt6_applications` counterparts, so we'll probably need hooks for those as well.

-------

However, assuming this package is meant to be used as follows:

```
#from PyQt5 import QtCore, QtWidgets

from pyqt5_plugins import import_it
QtCore = import_it('PyQt', 'QtCore')
QtWidgets = import_it('PyQt', 'QtWidgets')
```

this means that the user will need to add all used Qt-related imports (except `PyQtX.QtCore`, which is already handled by above hooks) to hiddenimports. I suppose with recent PyInstaller versions, we could implement bytecode scanner for two-argument `import_it()` calls where the first argument is `PyQt`, and convert those into hiddenimports for `PyQtX.QtModname`, but it seems like a lot of hassle with limited benefits. 

Might as well tell users to avoid using this wrapper and use PyQt5/PyQt6 directly...

-----

In addition, it looks like these packages are trying (or planning on trying) to overriding the Qt libraries and plugins provided by base PyQt package. For example, see contents of `site-packages\qt5_applications\Qt\bin` and `site-packages\qt5_applications\Qt\plugins` on Windows. That won't work with our Qt hooks at all, because it would likely require the regular PyQt5 hook to import the wrapper package before trying to determine plugins location.

Not that the override seems to (currently) work for the unfrozen version, either. The following test program:

```python
import os
import sys
import signal

from pyqt5_plugins import import_it
QtCore = import_it('PyQt', 'QtCore')
QtWidgets = import_it('PyQt', 'QtWidgets')
QtDesigner = import_it('PyQt', 'QtDesigner')

#from PyQt5 import QtCore, QtWidgets, QtDesigner


# e.g. finding the shared libs (dll/so) our python process loaded so far ...
def get_loaded_dlls():
    import psutil
    p = psutil.Process(os.getpid())
    return [dll.path for dll in p.memory_maps()]
        

app = QtWidgets.QApplication(sys.argv)

window = QtWidgets.QMainWindow()
window.setWindowTitle("Loaded DLL list")
window.resize(800, 600)
window.show()

scrollArea = QtWidgets.QScrollArea()
scrollArea.setWidgetResizable(True)
window.setCentralWidget(scrollArea)

textEdit = QtWidgets.QTextEdit()
textEdit.setReadOnly(True)
scrollArea.setWidget(textEdit)

print("PLUGINS PATH: %s" % (QtCore.QLibraryInfo.location(QtCore.QLibraryInfo.PluginsPath)))

textEdit.setText("\n".join(sorted(get_loaded_dlls())))

app.exec_()
```

reports that both Qt libraries and plugins are still being loaded from PyQt5 package (e.g., `site-packages\PyQt5\Qt5\bin\Qt5Core.dll`, `site-packages\PyQt5\Qt5\plugins\platforms\qwindows.dll`).

But if the ultimate idea is to replace/override the Qt binaries and plugins from PyQt5 package, then I think it'd be better to raise an error if use of `qt5_plugins` is detected, rather than handle metadata and hiddenimports and pretend that everything is fine, all while we end up silently collecting the original binaries from PyQt5 wheel.

PR #287, opened 2021-08-05
hooks: add hook for pynput

eric15342335:
Pynput [chooses](https://github.com/moses-palmer/pynput/blob/master/lib/pynput/_util/__init__.py#L49) different backends for different platforms, but users can specify via [environment variables](https://pynput.readthedocs.io/en/latest/#forcing-a-specific-backend) too.

I am not sure if this is right.


eric15342335:
or we can use alternative method like how pynput loads backend

bwoodsend:
Hmm, what we're [trying to with the matplotlib hook](https://github.com/pyinstaller/pyinstaller/pull/6024) (which also has a choice of backends) is [make the selection configurable](https://pyinstaller.readthedocs.io/en/stable/hooks-config.html#adding-an-option-to-the-hook). In this case it would be to have whatever backend pynput selects by default as the default then have a hook option to choose a backend (or possibly backends if needing more than one is likely?).

bwoodsend:
Actually looking at their available backends, this isn't really like matplotlib where you have more choices than you need, it's pretty much just one per OS meaning that your `collect_submodules()` approach is fine.

bwoodsend:
It does have per-platforn dependencies (`python-xlib` for Linux, `pyobjc-framework-Quartz` for macOS`) which will likely become warnings on other OSs because they're not installed.

src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-pynput.py
```diff
@@ -0,0 +1,14 @@
...
+# License (version 2.0 or later).
+#
+# The full license is available in LICENSE.GPL.txt, distributed with
+# this software.
+#
+# SPDX-License-Identifier: GPL-2.0-or-later
+# ------------------------------------------------------------------
+
+from PyInstaller.utils.hooks import collect_submodules
+hiddenimport = collect_submodules("pynput")
```
bwoodsend, line 14: It [fails testing](https://github.com/bwoodsend/pyinstaller-hooks-contrib/actions/runs/1103129291) because of this:
```suggestion
hiddenimports = collect_submodules("pynput")
```

bwoodsend (changes requested):

bwoodsend:
Here's [CI](https://github.com/bwoodsend/pyinstaller-hooks-contrib/actions/runs/1105666869) and [some more CI](https://github.com/bwoodsend/pyinstaller-hooks-contrib/actions/runs/1105713359) becuase Linux needs a virtual screen.

bwoodsend:
Cheers @eric15342335!

PR #288, opened 2021-08-07
hooks: add hook for pystray

eric15342335:
Pystray [chooses](https://github.com/moses-palmer/pystray/blob/master/lib/pystray/__init__.py#L22) different backends for different platforms. Note that if [this branch](https://github.com/moses-palmer/pystray/tree/feature-explicit-backends) get merged then we don't need this.
(This is the same)

[**CI**](https://github.com/eric15342335/pyinstaller-hooks-contrib/actions/runs/1107971245)

### DIFF
--- a/news/288.new.rst
+++ b/news/288.new.rst
@@ -1,0 +1,1 @@
+Add a hook for `pystray` to collect hidden imports.
--- a/news/288.new.rst
+++ b/news/288.new.rst
@@ -1,1 +1,1 @@
-Add a hook for `pystray` to collect hidden imports.
+Add a hook for ``pystray`` to collect hidden imports.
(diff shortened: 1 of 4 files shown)


news/288.new.rst
```diff
@@ -0,0 +1 @@
+Add a hook for `pystray` to collect hidden imports.
```
rokm, line 1:
```suggestion
Add a hook for ``pystray`` to collect hidden imports.
```

rokm:
> Note that if [this branch](https://github.com/moses-palmer/pystray/tree/feature-explicit-backends) get merged then we don't need this.

True, but having the hook will ensure compatibility with all versions of `pystray`. 

And as this is a relatively simple package, having a (potentially) redundant `collect_submodules` should not be a problem...

bwoodsend:
> Note that if this branch get merged then we don't need this.

Mind putting that in a comment in the hook to aid any future cleaups?

bwoodsend (approved):

Issue #140, opened 2021-07-03
What to do about scheduled CI/CD?


bwoodsend:
I've got two new issues regarding our weekly *do they still work* CI:

1.  [AppVeyor jobs are so long that they're timing out](https://ci.appveyor.com/project/PyInstaller/pyinstaller-hooks-contrib/builds/39846121). Each job has a 1 hour time limit and we've just started hitting that with the Python 3.6 builds and we're only a few minutes short for the others. 
  **Options**:
    * It says on their [pricing page](https://www.appveyor.com/pricing/) *All plans have maximum build job execution time of 60 minutes. For longer build times contact us.* which I suppose we could just do - it'd be the simplest option by far.
    * Split each job into two by dividing the requirements file. This feels rather like pointlessly jumping through hoops to me and it probably eats more CI time overall than one long job because tests for shared dependencies may run twice.
    * Abandon AppVeyor in favour of Github Actions [which has a 6 hour job limit](https://docs.github.com/en/actions/reference/usage-limits-billing-and-administration#usage-limits). 

    It's occurred to me that the *test with PyInstaller on PyPI* permutation is pointless as everything is pinned so that can go (changes in progress) which will half the total workflow time but do nothing to address the issue of overlong jobs.

2.   Our Travis schedule seems to have stopped triggering since travis-ci.org migrated to travic-ci.com two and a bit weeks ago. [Our last build on travis-ci.org ran 19 days ago](https://www.travis-ci.org/github/pyinstaller/pyinstaller-hooks-contrib/builds) and [there has been no new activity on travis-ci.com](https://www.travis-ci.com/github/pyinstaller/pyinstaller-hooks-contrib/builds). Looking at the [settings page](https://www.travis-ci.com/github/pyinstaller/pyinstaller-hooks-contrib/settings) the weekly scheduling had copied over but the next build was due next January so I've deleted and recreated that schedule rule and now it says *Scheduled in 22 minutes ago* but nothing appears to have been launched.
    I reckon we should just try waiting a week to see if a build appears before thinking about anything more drastic. They used to be unconfigurably triggered on Mondays so perhaps one will appear then.

rokm:
Hmm... well, I suppose it doesn't hurt to ask AppVeyor about longer build times. And if that's a no-go, migrate over to GitHub actions. Splitting jobs does sound like jumping through hoops, and should be avoided. 

[...]

PR #290, opened 2021-08-07
CI/CD: Migrate tests on TravisCI to Github Actions.

bwoodsend:
Following on from #140 - TravisCI is looking like more trouble than it's worth so I've moved it to Github to join macOS.

I've left the Windows CI on AppVeyor. We'll probably move it to Github too eventually but I'm reluctant to do it yet - I'm rather fond of AppVeyor at the moment.


bwoodsend:
I did a scaled down test run ([here](https://github.com/bwoodsend/pyinstaller-hooks-contrib/actions/runs/1108651746)) just to verify that I'd copied across all the Linux-only apt commands and setup.

Legorooj (approved):

rokm (approved):

PR #291, opened 2021-08-11
hooks: add hook for rtree (rokm)

eric15342335:
https://github.com/pyinstaller/pyinstaller/discussions/6060#discussioncomment-1142925

### DIFF
--- a/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-rtree.py
+++ b/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-rtree.py
@@ -10,7 +10,7 @@
 # SPDX-License-Identifier: GPL-2.0-or-later
 # ------------------------------------------------------------------
 
-from PyInstaller.utils.hooks import is_pure_conda
+from PyInstaller.compact import is_pure_conda
 
 if is_pure_conda:
     from PyInstaller.utils.hooks.conda import collect_dynamic_libs
--- a/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-rtree.py
+++ b/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-rtree.py
@@ -10,7 +10,7 @@
 # SPDX-License-Identifier: GPL-2.0-or-later
 # ------------------------------------------------------------------
 
-from PyInstaller.compat import is_pure_conda
+from PyInstaller.compat import is_conda
 from PyInstaller.utils.hooks import collect_dynamic_libs, conda
 
 binaries = collect_dynamic_libs('rtree', destdir='rtree/lib')
--- a/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-rtree.py
+++ b/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-rtree.py
@@ -11,10 +11,8 @@
 # ------------------------------------------------------------------
 
 from PyInstaller.compat import is_pure_conda
+from PyInstaller.utils.hooks import collect_dynamic_libs, conda
 
-if is_pure_conda:
-    from PyInstaller.utils.hooks.conda import collect_dynamic_libs
-    binaries = collect_dynamic_libs('rtree', dest='rtree/lib')
-else:
-    from PyInstaller.utils.hooks import collect_dynamic_libs
-    binaries = collect_dynamic_libs('rtree', destdir='rtree/lib')
+binaries = collect_dynamic_libs('rtree', destdir='rtree/lib')
+if not binaries and is_pure_conda:
+    binaries = conda.collect_dynamic_libs('libspatialindex', dest='rtree/lib', dependencies=False)
--- a/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-rtree.py
+++ b/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-rtree.py
@@ -1,5 +1,5 @@
 # ------------------------------------------------------------------
-# Copyright (c) 2020 PyInstaller Development Team.
+# Copyright (c) 2021 PyInstaller Development Team.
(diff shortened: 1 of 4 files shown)


eric15342335:
the workflow isn't passing.

src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-rtree.py
```diff
@@ -0,0 +1,20 @@
...
+# This file is distributed under the terms of the GNU General Public
+# License (version 2.0 or later).
+#
+# The full license is available in LICENSE.GPL.txt, distributed with
+# this software.
+#
+# SPDX-License-Identifier: GPL-2.0-or-later
+# ------------------------------------------------------------------
+
+from PyInstaller.utils.hooks import is_pure_conda
```
bwoodsend: This should be:
```suggestion
from PyInstaller.compat import is_pure_conda
```

src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-rtree.py
```diff
@@ -0,0 +1,20 @@
...
+# This file is distributed under the terms of the GNU General Public
+# License (version 2.0 or later).
+#
+# The full license is available in LICENSE.GPL.txt, distributed with
+# this software.
+#
+# SPDX-License-Identifier: GPL-2.0-or-later
+# ------------------------------------------------------------------
+
+from PyInstaller.compact import is_pure_conda
```
bwoodsend, line 13:
```suggestion
from PyInstaller.compat import is_pure_conda
```

bwoodsend (changes requested):

bwoodsend:
I notice the docs for `conda_support` are wrong in that they suggest to do what you did and import `is_pure_conda` from `PyInstaller.utils.hooks`.

eric15342335:
Yea there's what I came from

rokm (changes requested):
Hmm, I think this should be reworked a bit...

First, we probably want to call `PyInstaller.utils.hooks.collect_dynamic_libs('rtree')` in all cases, because the user might have installed an `rtree` wheel from PyPI (for whatever reason) within their conda environment (and if I recall correctly, `is_pure_conda` won't catch that, because it distinguishes only between pure conda env and virtualenv within conda env).

Second, have you checked the output of `binaries = PyInstaller.utils.hooks.conda.collect_dynamic_libs('rtree', dest='rtree/lib')` ? That's going to be ugly and will add unnecessary increase the size to the bundled program. 

I suppose the entries need to be filtered, so that only `spatialindex*.dll` files are collected. 

Alternatively, we could skip collect_dynamic_libs and try to glob for `os.path.join(sys.prefix, 'Library', 'bin', 'spatialindex*.dll')`.

Or perhaps the best approach would be to instead use `binaries = PyInstaller.utils.hooks.conda.collect_dynamic_libs('libspatialindex', dest='rtree/lib', dependencies=False)`

eric15342335:
I made the history worser

bwoodsend:
> Or perhaps the best approach would be to instead use binaries = PyInstaller.utils.hooks.conda.collect_dynamic_libs('libspatialindex', dest='rtree/lib', dependencies=False)

I imagine this would be best. I'm having second thoughts about that default of `dependencies=True`.

bwoodsend:
> Yea that's where I came from, maybe a docs fix?

Yes please!

rokm:
Also, we should probably enter the conda-specific lookup only if `binaries` is still empty after the preceding call to `PyInstaller.utils.hooks.collect_dynamic_libs('rtree')`.

src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-rtree.py
```diff
@@ -11,10 +11,8 @@
...
+from PyInstaller.utils.hooks import collect_dynamic_libs, conda
 
-if is_pure_conda:
-    from PyInstaller.utils.hooks.conda import collect_dynamic_libs
-    binaries = collect_dynamic_libs('rtree', dest='rtree/lib')
-else:
-    from PyInstaller.utils.hooks import collect_dynamic_libs
-    binaries = collect_dynamic_libs('rtree', destdir='rtree/lib')
+binaries = collect_dynamic_libs('rtree', destdir='rtree/lib')
+if not binaries and is_pure_conda:
```
rokm, line 17: Is there a reason for using `is_pure_conda` instead of just `is_conda`?

bwoodsend: That should be `is_pure_conda`. The `conda_support` module doesn't exist if the user is running venv inside a conda environment.

eric15342335: I thought probably this
https://github.com/pyinstaller/pyinstaller/blob/develop/PyInstaller/compat.py#L165

bwoodsend: Yeah, `rtree` can't be a conda distribution unless `is_pure_conda` is true.

src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-rtree.py
```diff
@@ -0,0 +1,18 @@
+# ------------------------------------------------------------------
+# Copyright (c) 2020 PyInstaller Development Team.
```
rokm, line 2: Final nitpick: it's 2021 (even though these days time might appear to have frozen).

eric15342335: holy i copied the wrong one

rokm (approved):

PR #293, opened 2021-08-19
add cv2 config file

eric15342335:


eric15342335:
Will this work? I haven't tried this before

rokm:
Assuming this is targeting user-compiled OpenCV + python bindings (or official Windows OpenCV builds from opencv.org), the problems with that one run much deeper.

Hint: unlike in the PyPI wheel, `cv2` is not a package (does not have a `__path__`), nor does its `__file__` (at the point when you can observe it) point to where you think it does (i.e., the loader shim that's located next to those "config" scripts). And the config scripts are worthless anyway, because they assume paths that would take you outside of the PyInstaller-frozen bundle...

eric15342335:
hmmm, seems this is useless anyways...

PR #295, opened 2021-08-21
update CLR hook to set correct path for newer pythonnet packages

wmmc88:


Legorooj (changes requested):
4 spaces for indentation, not tabs, please.

Legorooj (commented):
Looks ok, would like @bwoodsend to sign off as well though.

bwoodsend:
Hmm, to me this raises several questions, all applicable to the original hook rather than these changes.

1. Could all that site-packages iterating be replaced with something using `importlib.metadata.files("pythonnet")` or similar?
2. Can we get rid of [these lines](https://github.com/pyinstaller/pyinstaller-hooks-contrib/pull/295/files#diff-ff4f0df1061b70b893c3f12af1c8697a51d8bffba6e5e8bf5257ba743cc7d9f3R32-R33)? I'd have thought that in the unlikely event that there is a `Python.net.dll` in `PATH`, it'll either be the one we're about to collect in the lines below (i.e. it's redundant) or it's another one and therefore the wrong one.

If your answers to the above are no and no then I approve.

wmmc88:
> Hmm, to me this raises several questions, all applicable to the original hook rather than these changes.
> 
> 1. Could all that site-packages iterating be replaced with something using `importlib.metadata.files("pythonnet")` or similar?
> 2. Can we get rid of [these lines](https://github.com/pyinstaller/pyinstaller-hooks-contrib/pull/295/files#diff-ff4f0df1061b70b893c3f12af1c8697a51d8bffba6e5e8bf5257ba743cc7d9f3R32-R33)? I'd have thought that in the unlikely event that there is a `Python.net.dll` in `PATH`, it'll either be the one we're about to collect in the lines below (i.e. it's redundant) or it's another one and therefore the wrong one.
> 
> If your answers to the above are no and no then I approve.

1. they could, but `importlib.metadata` is a python3.8+ module. Not sure what python versions this contrib package is supposed to support, but i guess there also exists the `import_lib.metadata` [backport package](https://importlib-metadata.readthedocs.io/en/latest/) that can be used.
2. I'm not sure if maybe one of the older versions of pythonnet packaged it differently, hence those lines. I'd be hesitant to remove it in case it breaks someone's build (especially since there doesn't seem to be any tests for this hook). I could just make it the last ditch effort to find the dependency, instead of the first thing that's tried.

Speaking of tests, I wouldn't mind writing a quick test to go with this, but I'm not sure if the repo has some way to easily test different versions of the package working with the hook(ie. pythonnet 2 and pythonnet 3)

bwoodsend:
> they could, but importlib.metadata is a python3.8+ module. Not sure what python versions this contrib package is supposed to support, but i guess there also exists the import_lib.metadata backport package that can be used.

The `importlib-metadata` backport [is already a dependency of PyInstaller for older Pythons](https://github.com/pyinstaller/pyinstaller/blob/34b443a547fddabf06bf15889e56b430d31da5a5/setup.cfg#L65) so yes you can use that.
> I could just make it the last ditch effort to find the dependency, instead of the first thing that's tried.

Sounds reasonable. Can you do that then please.

> Speaking of tests, I wouldn't mind writing a quick test to go with this, but I'm not sure if the repo has some way to easily test different versions of the package working with the hook(ie. pythonnet 2 and pythonnet 3).

See [our oneshot CI](https://github.com/bwoodsend/pyinstaller-hooks-contrib/tree/readme#run-the-test-on-cicd).

src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-clr.py
```diff
@@ -18,23 +18,34 @@
 
 
 import ctypes.util
-from PyInstaller.compat import is_win, getsitepackages
-from os.path import join, exists
+from PyInstaller.compat import is_win
+from PyInstaller.utils.hooks import logger
```
bwoodsend, line 22: Take the logger from `PyInstaller.log` please.
```suggestion
from PyInstaller.log import logger
```
It's not supposed to be available in `PyInstaller.utils.hooks` - rather some functions in there use the logger and there's no proper namespace hiding so it looks like public API when it isn't.

bwoodsend (commented):
I doubt anyone will notice those logger warnings but that doesn't really matter. Did you get the oneshot CI working? If not, I can trigger it.

wmmc88:
> I doubt anyone will notice those logger warnings but that doesn't really matter. Did you get the oneshot CI working? If not, I can trigger it.

Still writing a test atm

bwoodsend:
How do you want to deal with Linux and mac? I understand that despite officially being cross platform the .Net framework is very much a Windows thing. That said, the test passes on Linux if I remove the `if is_win` guard in the hook so, unless macOS is more complex, it looks like making it cross platform would be trivial.

wmmc88:
> How do you want to deal with Linux and mac? I understand that despite officially being cross platform the .Net framework is very much a Windows thing. That said, the test passes on Linux if I remove the `if is_win` guard in the hook so, unless macOS is more complex, it looks like making it cross platform would be trivial.

Yeah so ".Net Framework" is only windows, but ".Net Core" is officially cross-platform. I wrote the test using .Net Core so it _should_ just work. Might need to make sure the CI runner has .net tho. Let me try it out...

wmmc88:
> > How do you want to deal with Linux and mac? I understand that despite officially being cross platform the .Net framework is very much a Windows thing. That said, the test passes on Linux if I remove the `if is_win` guard in the hook so, unless macOS is more complex, it looks like making it cross platform would be trivial.
> 
> Yeah so ".Net Framework" is only windows, but ".Net Core" is officially cross-platform. I wrote the test using .Net Core so it _should_ just work. Might need to make sure the CI runner has .net tho. Let me try it out...

https://github.com/wmmc88/pyinstaller-hooks-contrib/runs/3422624145?check_suite_focus=true
works on mac and windows. looks like the ubuntu image doesn't have dotnet by default. I could pull it in using https://github.com/actions/setup-dotnet

wmmc88:
Latest run:
https://github.com/wmmc88/pyinstaller-hooks-contrib/actions/runs/1166986318

bwoodsend:
Could the test be made to run with pythonnet 2.x? Other than that, looks good.

wmmc88:
Ok I made a separate test to run for pythonnet 2.x.x, but this test should only succeed on python <3.9 (due to a pythonnet bug preventing python 3.9 support in the latest v2 release) and only on windows platforms (since pythonnet 2 only supports .net framework and that is windows only).

You can see the new 2.x test succeeding here: https://github.com/wmmc88/pyinstaller-hooks-contrib/runs/3423816991?check_suite_focus=true

I had to however comment out the v3 test on that branch i ran it on because of a bug I think i found in the `requires` decorator. As you can see in the v[2](https://github.com/wmmc88/pyinstaller-hooks-contrib/actions/runs/1167178251) and [v3](https://github.com/wmmc88/pyinstaller-hooks-contrib/runs/3423893464?check_suite_focus=true) tests, the requires for the other version is not being marked as skipped. It is failing due to an uncaught `pkg_resources.VersionConflict` exception.[

This line in Pyinstaller](https://github.com/pyinstaller/pyinstaller/blob/23ac958f58624d21c2c08c6b1dc0491d1e00e7eb/PyInstaller/utils/tests.py#L121) should probably say `except (pkg_resources.DistributionNotFound, pkg_resources.VersionConflict):`.

bwoodsend:
> This line in Pyinstaller should probably say except (pkg_resources.DistributionNotFound, pkg_resources.VersionConflict):.

Yeah, you're right. Or possibly it could be the base class `ResolutionError`.

bwoodsend:
So if the pythonnet 2 test is only applicable to windows, can we mark as such and the make the entry in the requirements file Windows only too. Then it's probably easiest to remove the *install pythonnet 3 from git* line and replace it with a comment saying *pythonnet 3 (when released) will be cross platform*. That way, PyUp will automatically notify us when pythonnet 3 is on PyPI and I can then enable pythonnet 3 testing on all platforms.

PR #300, opened 2021-08-23
hooks: add uvicorn

imba-tjd:
I have seen https://github.com/pyinstaller/pyinstaller/pull/4664
But I think it's OK to include the whole uvicorn, because it's pure Python.


src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-uvicorn.py
```diff
@@ -0,0 +1,15 @@
+# ------------------------------------------------------------------
+# Copyright (c) 2020 PyInstaller Development Team.
```
rokm, line 2: 2021

imba-tjd: done

bwoodsend:
Ughh I hate libraries that do this. Having both dynamical imports and optional dependencies means that we simultaneously must and must not use `collect_submodules()`. And our tests will always be either hopelessly inadequate in that they provide almost zero coverage or have to brute force iterate through all the permutations and take hours like our `scikit-image` tests.

bwoodsend:
Thanks @imba-tjd.

Issue #285, opened 2017-07-06
"FileNotFoundError: [Errno 2]" when creating exe with langdetect


yucongo:
# applangdetect.py
    import langdetect

    print('test', langdetect.detect('this is a test'))
    # end of applangdetect.py

pyinstaller applangdetect.py

applangdetect.exe

    Traceback (most recent call last):
      File "applangdetect.py", line 1, in <module>
      File "<frozen importlib._bootstrap>", line 2237, in _find_and_load
      File "<frozen importlib._bootstrap>", line 2226, in _find_and_load_unlocked
      File "<frozen importlib._bootstrap>", line 1200, in _load_unlocked
      File "<frozen importlib._bootstrap>", line 1129, in _exec
      File "d:\python34\lib\site-packages\PyInstaller\loader\pyimod03_importers.py",
     line 631, in exec_module
        exec(bytecode, module.__dict__)
      File "site-packages\langdetect\__init__.py", line 1, in <module>
      File "<frozen importlib._bootstrap>", line 2237, in _find_and_load
      File "<frozen importlib._bootstrap>", line 2226, in _find_and_load_unlocked
      File "<frozen importlib._bootstrap>", line 1200, in _load_unlocked
      File "<frozen importlib._bootstrap>", line 1129, in _exec
      File "d:\python34\lib\site-packages\PyInstaller\loader\pyimod03_importers.py",
     line 631, in exec_module
        exec(bytecode, module.__dict__)
      File "site-packages\langdetect\detector_factory.py", line 10, in <module>
      File "<frozen importlib._bootstrap>", line 2237, in _find_and_load
      File "<frozen importlib._bootstrap>", line 2226, in _find_and_load_unlocked
      File "<frozen importlib._bootstrap>", line 1200, in _load_unlocked
      File "<frozen importlib._bootstrap>", line 1129, in _exec
      File "d:\python34\lib\site-packages\PyInstaller\loader\pyimod03_importers.py",
     line 631, in exec_module
        exec(bytecode, module.__dict__)
      File "site-packages\langdetect\detector.py", line 9, in <module>
      File "<frozen importlib._bootstrap>", line 2237, in _find_and_load
      File "<frozen importlib._bootstrap>", line 2226, in _find_and_load_unlocked
      File "<frozen importlib._bootstrap>", line 1200, in _load_unlocked
      File "<frozen importlib._bootstrap>", line 1129, in _exec
[...]

PR #302, opened 2021-08-24
hooks: add hook for langdetect and swagger_spec_validator

eric15342335:
#285 and #296
I seems can not run the workflow
sshot-001

### DIFF
--- a/news/285.new.rst
+++ b/news/285.new.rst
@@ -1,0 +1,1 @@
+Add a hook for ``langdetect`` to collect data files.
--- a/news/296.new.rst
+++ b/news/296.new.rst
@@ -1,0 +1,1 @@
+Add a hook for ``swagger_spec_validator`` to collect data files.
--- a/requirements-test-libraries.txt
+++ b/requirements-test-libraries.txt
@@ -38,6 +38,7 @@ sentry-sdk==0.19.3
 git+https://github.com/shotgunsoftware/python-api.git@v3.2.6
 spacy==3.0.6
 srsly==2.4.1
+swagger-spec-validator==2.7.3
 thinc==8.0.4
 timezonefinder==5.2.0
 Unidecode==1.1.1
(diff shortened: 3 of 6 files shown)


bwoodsend:
>  I seems can not run the workflow

This is rather odd. The actions for both [the master repo](https://github.com/pyinstaller/pyinstaller-hooks-contrib/actions/) and [my fork](https://github.com/bwoodsend/pyinstaller-hooks-contrib/actions/) are behaving normally so it appears to be specific to your fork but your master branch is even with ours so I don't see what the difference could be. In [eric15342335/pyinstaller-hooks-contrib/settings/actions](https://github.com/eric15342335/pyinstaller-hooks-contrib/settings/actions), do you have *Allow all actions* selected? That's what we have.

bwoodsend (changes requested):
We need news entries for each one. Use the issue numbers instead of the PR numbers in the news entries' filenames.

bwoodsend:
I'd prefer that each news entry was added in the same commit as the hook and test rather than an extra *add news item* commit but it's not that important. I've triggered CI from my own fork for now. https://github.com/bwoodsend/pyinstaller-hooks-contrib/actions/runs/1162353622

bwoodsend:
@eric15342335 I'd like to add you to the PyInstaller team and give you basic triaging permissions. Basically you'd be able to do what I do here - make sure new pull requests make sense and have tests and news entries. Then I'll eventually give you push access so that you can start merging yourself too. You'll also be able to do the same on the main PyInstaller repo too. Sound OK?

eric15342335:
Sounds pretty good, thank you!

bwoodsend:
OK, let me work out who has permission to grant other people permissions...

bwoodsend:
In theory, you should have received an invite. Could you also [send me an email](mailto:bwoodsend@gmail.com?Subject=Joining%20PyInstaller&Body=Hello%2C%20I%27m%20Eric.) so that I can send you a link to the Discord channel we use to talk to each other.

PR #312, opened 2021-09-02
Per-platform imports for platformdirs

jalaziz:
Only import the platformdirs submodule for the platform the binary is
being built against.

### DIFF
--- a/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-platformdirs.py
+++ b/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-platformdirs.py
@@ -10,7 +10,18 @@
 # SPDX-License-Identifier: GPL-2.0-or-later
 # ------------------------------------------------------------------
 
-from PyInstaller.utils.hooks import collect_submodules
+from PyInstaller.compat import is_darwin, is_win
+
+modules = ["platformdirs"]
 
 # platfromdirs contains dynamically loaded per-platform submodules.
-hiddenimports = collect_submodules("platformdirs")
+if is_darwin:
+    modules.append("platformdirs.macos")
+elif is_win:
+    modules.append("platformdirs.windows")
+else:
+    # default to unix for all other platforms
+    # this includes unix, cygwin, and msys2
+    modules.append("platformdirs.unix")
+
+hiddenimports = modules


gaborbernat (approved):

bwoodsend:
Are you sure our definitions of `is_darwin`, `is_unix`, `is_win` match those of platformdirs? In particular Cygwin and MSYS2 are not considered Unix by PyInstaller despite their being more POSIX compliant than some Linux distributions and FreeBSD typically floats somewhere between macOS and Linux.

bwoodsend:
I just looked it up myself. So comparing [your platform selection](https://github.com/platformdirs/platformdirs/blob/36f17d62269f5715b23efeb7ba9e5195df5ae750/src/platformdirs/__init__.py#L18-L28) with our definitions of [is_win](https://github.com/pyinstaller/pyinstaller/blob/42046c04bb8f814d7fa0e7e0c9e8b4ce19ec2b85/PyInstaller/compat.py#L39), [is_darwin](https://github.com/pyinstaller/pyinstaller/blob/42046c04bb8f814d7fa0e7e0c9e8b4ce19ec2b85/PyInstaller/compat.py#L43) and [is_unix](https://github.com/pyinstaller/pyinstaller/blob/42046c04bb8f814d7fa0e7e0c9e8b4ce19ec2b85/PyInstaller/compat.py#L55), the only discrepancy I see is that you consider Cygin and MSYS2 to be Unix but we don't consider them to be either Unix or Windows.

jalaziz:
> I just looked it up myself. So comparing [your platform selection](https://github.com/platformdirs/platformdirs/blob/36f17d62269f5715b23efeb7ba9e5195df5ae750/src/platformdirs/__init__.py#L18-L28) with our definitions of [is_win](https://github.com/pyinstaller/pyinstaller/blob/42046c04bb8f814d7fa0e7e0c9e8b4ce19ec2b85/PyInstaller/compat.py#L39), [is_darwin](https://github.com/pyinstaller/pyinstaller/blob/42046c04bb8f814d7fa0e7e0c9e8b4ce19ec2b85/PyInstaller/compat.py#L43) and [is_unix](https://github.com/pyinstaller/pyinstaller/blob/42046c04bb8f814d7fa0e7e0c9e8b4ce19ec2b85/PyInstaller/compat.py#L55), the only discrepancy I see (minus our not supporting Android) is that you consider Cygin and MSYS2 to be Unix but we don't consider them to be either Unix or Windows.

Great catch. Will fix it.

jalaziz:
@bwoodsend I've adjusted the `if`-block to better match the `platformdirs` logic.

bwoodsend:
Yeah that should do it. As we never released the old version of this hook, there's no need for a news item.

CI run is [here](https://github.com/bwoodsend/pyinstaller-hooks-contrib/actions/runs/1195890220). If it passes then we're good to go

bwoodsend:
Thanks @jalaziz!

PR #324, opened 2021-09-25
hooks: add hook for pymediainfo

rokm:
Collects bundled `MediaInfo` shared library that is available in macOS and Windows wheels. On other OSes, it attempts to collect
the system-installed `MediaInfo` shared library.

The tests are currently enabled only for macOS and Windows, where we want to check if bundled shared library is properly collected. If we wanted the test to run on Linux, we'd need to install `libmediainfo0v5` via `apt`, but that would be self-defeating, because once system copy is available, the frozen test would pass if the library was collected or not.

### DIFF
--- a/news/324.new.rst
+++ b/news/324.new.rst
@@ -1,0 +1,1 @@
+Add hook for ``pymediainfo``.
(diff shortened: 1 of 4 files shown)


rokm:
Oneshot test: https://github.com/rokm/pyinstaller-hooks-contrib/actions/runs/1273692690

bwoodsend:
> If we wanted the test to run on Linux, we'd need to install libmediainfo0v5 via apt, but that would be self-defeating, because once system copy is available, the frozen test would pass if the library was collected or not.

Well we could rip up all the floorboards, re jiggle the entire test suite so that building happens in one docker container and testing in a second. Strangely, I'd rather not...

bwoodsend (approved):

PR #333, opened 2021-10-13
Update hook-apscheduler.py

iamliamc:
Update hook-apscheduler copy_metadata with recursive=True

Based on work with package author on this issue: 
https://github.com/agronholm/apscheduler/issues/556

### DIFF
--- a/src/_pyinstaller_hooks_contrib/tests/test_libraries.py
+++ b/src/_pyinstaller_hooks_contrib/tests/test_libraries.py
@@ -78,14 +78,10 @@ def test_apscheduler(pyi_builder):
         import datetime as dt
         from apscheduler.schedulers.asyncio import AsyncIOScheduler
         from apscheduler.triggers.interval import IntervalTrigger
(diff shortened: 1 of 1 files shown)


bwoodsend:
> And it feels a bit heavy handed...

It's not too heavy handed IMO. Maybe drop the 60 seconds to 30. Our CI is very much geared towards only running specific tests when they need to be ran so the fact that this test will take a bit longer than most won't get in the way. But either way, we still need the [news entry](https://github.com/pyinstaller/pyinstaller-hooks-contrib/blob/master/news/README.txt) and as [agronholm says](https://github.com/pyinstaller/pyinstaller-hooks-contrib/pull/333#issuecomment-942118926) wrap the whole hook in a `if is_module_satifies("apscheduler < 4"):` - at the very least, this serves as a signal to us PyInstaller devs when the hook becomes obsolete.

bwoodsend:
Nice work @iamliamc!

Issue #332, opened 2021-10-13
certifi runtime hook interferes with ssl.load_default_certs()


LincolnPuzey:
**Describe the bug**
The certifi runtime hook which was added in https://github.com/pyinstaller/pyinstaller/pull/3952 sets the `SSL_CERT_FILE` environment variable to the path  to cerifi's `cacert.pem`.

This causes all the certificates in `cacert.pem` to be loaded by `ssl.load_default_certs()` and therefore by included in the SSLContext created by `ssl.create_default_context()`

**To Reproduce**

My script `print_certs.py`
```
import ssl

import requests
import certifi

print(certifi.where())
context = ssl.create_default_context()
print(context.cert_store_stats())
```

output when running from source in venv with requests/certifi installed
```
PS C:\Users\Lincoln\cert_test> .\.venv\Scripts\python .\print_certs.py
C:\Users\Lincoln\cert_test\.venv\lib\site-packages\certifi\cacert.pem
{'x509': 27, 'crl': 0, 'x509_ca': 25}
```

Output when bundled with pyinstaller (all default options)
```
PS C:\Users\Lincoln\cert_test> .\dist\print_certs\print_certs.exe
C:\Users\Lincoln\cert_test\dist\print_certs\certifi\cacert.pem
{'x509': 148, 'crl': 0, 'x509_ca': 146}
```

**Expected behavior**
I would expect the output of `cert_store_stats()` for bundled to be the same as non-bundled

**Desktop:**
 - OS: Windows
 - Python Version: 3.7.9
 - Version of `pyinstaller-hooks-contrib`: 2021.3
 - Version of PyInstaller 4.5.1

**Proposed Solution**
I think the certifi runtime hook should be removed.
* The `SSL_CERT_FILE` env variable is not required for certifi to function
* There is already a stdhook for certifi to ensure its data files are included https://github.com/pyinstaller/pyinstaller-hooks-contrib/blob/master/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-certifi.py

bwoodsend:
I don't know anything about certifi so I can't say much other than that I would also expect PyInstaller's behaviour to mirror the unfrozen behaviour. I'd really like to know what the rational behind pyinstaller/pyinstaller#3952 though. @ktong Are you still on Github? Can you comment?

LincolnPuzey:
certifi provides root certificates in the form of the `cacert.pem` file. Quoting the certifi README;

[...]

PR #335, opened 2021-10-19
[Issue #332] Delete certifi runtime hook

LincolnPuzey:
This hook is being deleted because it is not required for certifi to function and it introduces inconsistent behavior between frozen/unfrozen code.

See issue #332 for details.


LincolnPuzey:
closes #332

LincolnPuzey:
Also I think I need to remove the reference to this hook in `src/_pyinstaller_hooks_contrib/hooks/rthooks.dat`

PR #336, opened 2021-10-19
Include markdown metadata to support entry-points

ovidner:
The markdown package uses entry-points to allow short names for built-in extensions, such as `extra` or `toc`. The hook has been updated so that the necessary package metadata is included in the build, enabling the same behavior as the unfrozen package.

This would resolve problems like https://github.com/Python-Markdown/markdown/issues/884.

### DIFF
--- a/news/336.update.rst
+++ b/news/336.update.rst
@@ -1,0 +1,1 @@
+Update ``markdown`` hook to include package metadata, enabling the use of short names for built-in extensions, such as ``extra`` or ``toc``.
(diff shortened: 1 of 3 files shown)


bwoodsend:
CI/CD run is [here](https://github.com/bwoodsend/pyinstaller-hooks-contrib/actions/runs/1363117643).

ovidner:
> 
> 
> I hope that you are aware though that other providers of mardown's entry point will still not be collected.

Ah yes, that's true. I guess that it could be even better to use `collect_entry_point("markdown.extensions")`, considering that. But for the time being, I think this resolves the biggest issue. At least it's a fix for my own use-case 😁

PR #348, opened 2021-11-10
Add a hook for ffpyplayer 

eric15342335:
https://stackoverflow.com/questions/69884761/ffpyplayer-importerror-dll-load-failed-while-importing-player-the-specified

See [here](https://github.com/matham/ffpyplayer/blob/ec3af54c9b45828bdf6691b064c6f4f9e46169f5/ffpyplayer/__init__.py#L27-L29)

### DIFF
--- a/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-ffpyplayer.py
+++ b/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-ffpyplayer.py
@@ -1,0 +1,19 @@
+# ------------------------------------------------------------------
+# Copyright (c) 2021 PyInstaller Development Team.
+#
+# This file is distributed under the terms of the GNU General Public
+# License (version 2.0 or later).
+#
+# The full license is available in LICENSE.GPL.txt, distributed with
+# this software.
+#
+# SPDX-License-Identifier: GPL-2.0-or-later
+# ------------------------------------------------------------------
+
+from PyInstaller.utils.hooks import eval_statement, collect_submodules
+
+hiddenimports = collect_submodules("ffpyplayer")
+binaries = []
+# ffpyplayer has an internal variable tells us where the libraries it was using
+for bin in eval_statement("import ffpyplayer; print(ffpyplayer.dep_bins)"):
+    binaries += [(bin, '.')]
--- a/news/348.new.rst
+++ b/news/348.new.rst
@@ -1,0 +1,1 @@
+Add a hook for ``ffpyplayer``.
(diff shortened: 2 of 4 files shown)


eric15342335:
ok i'll add new line

Legorooj (commented):
Looks good, though you may want to change the commit message haha.

eric15342335:
> 
> 
> Looks good, though you may want to change the commit message haha.

sorry for that

src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-ffpyplayer.py
```diff
@@ -0,0 +1,19 @@
...
+#
+# SPDX-License-Identifier: GPL-2.0-or-later
+# ------------------------------------------------------------------
+
+from PyInstaller.utils.hooks import eval_statement, collect_submodules
+
+hiddenimports = collect_submodules("ffpyplayer")
+binaries = []
+# ffpyplayer has an internal variable tells us where the libraries it was using
+for bin in eval_statement("import ffpyplayer; print(ffpyplayer.dep_bins)"):
```
rokm, line 18: I think long-term it would make sense to use `get_module_attribute` here. At the moment, though, 4.x returns string-ified value, while 5.x returns native-type version (presumably a list), so explicitly handling both might not be worth the hassle.

src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-ffpyplayer.py
```diff
@@ -0,0 +1,19 @@
...
+#
+# SPDX-License-Identifier: GPL-2.0-or-later
+# ------------------------------------------------------------------
+
+from PyInstaller.utils.hooks import get_module_attribute, collect_submodules
+
+hiddenimports = collect_submodules("ffpyplayer")
+binaries = []
+# ffpyplayer has an internal variable tells us where the libraries it was using
+for bin in get_module_attribute("ffpyplayer", "dep_bins"):
```
rokm, line 18: Ah sorry, looks like my previous comment regarding `get_module_attribute` was taken as request for change. The problem is that with PyInstaller 4.x, get_module_attribute returns string-ified version of attribute value, i.e.:

```
>>> from PyInstaller.utils.hooks import get_module_attribute
>>> get_module_attribute('ffpyplayer', 'dep_bins')
'[]'
```

So this needs to be `eval`d to get the list.

In PyInstaller 5.x we switched to new process isolation framework that returns proper native types, so:

```
>>> from PyInstaller.utils.hooks import get_module_attribute
>>> get_module_attribute('ffpyplayer', 'dep_bins')
[]
```

(i.e., the list is returned and can be used directly).

So this would require different handling based on PyInstaller version (that's why my previous comment was meant more as a side note, and mentioned the hassle of handling both).

bwoodsend (approved):
And I see [you've done CI](https://github.com/eric15342335/pyinstaller-hooks-contrib/actions/runs/1444019065).

PR #350, opened 2021-11-18
fix missing changelog

imba-tjd:
I didn't notice the naming convention


bwoodsend:
No, you were right the first time. We write changelog entries in those fragments then when we do a release we run towncrier to move all the fragments into the changelog itself. If everyone was editing the changelog.rst file directly then we'd be getting tons of merge conflicts and it would be a nightmare.

rokm:
I think the problem is that the news fragment was incorrectly named, and we don't have a validator set up here...

bwoodsend:
Ohh, you're right. My mistake.

PR #351, opened 2021-11-18
CI/CD: Add towncrier news fragment validation workflow.

bwoodsend:
This is mostly just copied from PyInstaller's repo. If we find ourselves having to copy updates between the two then we could shunt it into a custom action later.


.github/workflows/validate-new-news.yml
```diff
@@ -0,0 +1,38 @@
...
+      - name: Setup Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.7
+
+      - name: Validate new news items
+        run: |
+          git diff --name-status origin/${{ github.base_ref }}
+          git diff --name-status origin/${{ github.base_ref }}\
+          | python scripts/verify-news-fragments.py
```
rokm, line 32: Doesn't the script work without the diffs now?

bwoodsend: Ughh, well remembered...

rokm (approved):

PR #352, opened 2021-11-18
Include metadata for pytorch

MoetaYuko:
Some 3rd-party packages (e.g., skorch)  requires pkg_resources.get_distribution()
of pytorch.

Example usage: https://github.com/skorch-dev/skorch/blob/9ac71e2acf8b814b65ee84ee6696e7441737e820/skorch/__init__.py#L29

### DIFF
--- a/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-torch.py
+++ b/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-torch.py
@@ -10,6 +10,6 @@
 # SPDX-License-Identifier: GPL-2.0-or-later
 # ------------------------------------------------------------------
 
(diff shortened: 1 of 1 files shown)


bwoodsend:
If it's another package that require pytorch's metadata then the hook should be for that package - not pytorch itself which doesn't need it.

But PyInstaller should be able to detect `pkg_resources.get_distribution('torch').version`. What version of PyInstaller are you using?

MoetaYuko:
> But PyInstaller should be able to detect `pkg_resources.get_distribution('torch').version` automatically. What version of PyInstaller are you using?

I was using 4.2 and this patch fixed the `pkg_resources.get_distribution('torch').version` error. Just tested with 4.7 and the error is gone, closing.

PR #354, opened 2021-11-25
Fixed an issue for hook-cv2 when extra config files were not found

asenyaev:
There are new hard dependencies in the new release of OpenCV python packages (4.5.4.60) on extra config files and modules.

This PR fixes following issues https://github.com/opencv/opencv-python/issues/569 and https://github.com/pyinstaller/pyinstaller-hooks-contrib/issues/110


asenyaev:
@bwoodsend please, take a look.

rokm (changes requested):
Ah, nice, opencv-python 4.5.4.60 removed the module substitution trickery, so indeed the only remaining problem are the missing .py files.

That said:
1) this is not a Windows-specific problem (i.e., the .py files should be collected on all platforms)
2) while temporarily modifying `PY_DYLIB_PATTERNS` is a clever way of tricking `collect_dynamic_libs` into collecting .py files, having the .py files collected as binaries will break on macOS. 

The .py files should be collected as data files instead, for example using `collect_data_files` function:
```python

from PyInstaller.utils.hooks import collect_dynamic_libs, collect_data_files

# existing binaries handling

# OpenCV loader from v4.5.3.58 on requires config files and submodules' __init__.py files 
# to be collected in their source form.
datas = collect_data_files('cv2', include_py_files=True, includes=['**\*.py'])
```
(by also specifying include patterns, we ensure that extension module does not get duplicated as data file).

asenyaev:
> Ah, nice, opencv-python 4.5.4.60 removed the module substitution trickery, so indeed the only remaining problem are the missing .py files.
> 
> That said:
> 
> 1. this is not a Windows-specific problem (i.e., the .py files should be collected on all platforms)
> 2. while temporarily modifying `PY_DYLIB_PATTERNS` is a clever way of tricking `collect_dynamic_libs` into collecting .py files, having the .py files collected as binaries will break on macOS.
> 
> The .py files should be collected as data files instead, for example using `collect_data_files` function:
> 
> ```python
> from PyInstaller.utils.hooks import collect_dynamic_libs, collect_data_files
> 
> # existing binaries handling
> 
> # OpenCV loader from v4.5.3.58 on requires config files and submodules' __init__.py files 
> # to be collected in their source form.
> datas = collect_data_files('cv2', include_py_files=True, includes=['**\*.py'])
> ```
> 
> (by also specifying include patterns, we ensure that extension module does not get duplicated as data file).

@rokm, thanks for such informative comment! I've changed the collecting `*.py` files using `collect_data_files` function.

src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-cv2.py
```diff
@@ -31,3 +31,5 @@
     # Include any DLLs from site-packages/cv2 (opencv_videoio_ffmpeg*.dll
     # can be found there in the PyPI version)
     binaries += collect_dynamic_libs('cv2')
+    # Need to include extra config files and modules
+    extra_data = collect_data_files('cv2', include_py_files=True, includes=['**\*.py'])
```
bwoodsend, line 35: Surely it should be `includes=['**/*.py']`?

rokm: > Surely it should be `includes=['**/*.py']`?

Uh, right, my bad.

And in addition, it should also be `datas` and not `extra_data`. And it should be outside the if block to be OS-agnostic...

asenyaev: Thanks for the hint!

rokm:
Alright, I think the only thing missing is a [news entry](https://github.com/pyinstaller/pyinstaller-hooks-contrib/blob/master/news/README.txt), then this should be good to go.

asenyaev:
@rokm I've added it. I'm sorry for such an unprepared pull request.

rokm:
> @rokm I've added it. I'm sorry for such an unprepared pull request.

Don't worry about it, we've all gone through this at least once...

The news fragment should inform the end-users what changed and why, so I think the message could be improved as: `Update ``cv2`` hook to collect extra config files and modules for compatibility with OpenCV 4.5.4.60.

I just noticed that we currently don't have even a basic import test for `cv2`; can I ask you to take a look at [`src/_pyinstaller_hooks_contrib/tests/test_libraries.py`](https://github.com/pyinstaller/pyinstaller-hooks-contrib/blob/master/src/_pyinstaller_hooks_contrib/tests/test_libraries.py) and add one at the end of the file? Just a basic `import cv2` test should suffice, so we can verify on the CI that the latest OpenCV version works now on all OSes.

asenyaev:
@rokm, I've updated a message accordingly to your template. In addition, I've added a test for `cv2` package in the end of `src/_pyinstaller_hooks_contrib/tests/test_libraries.py` file. 
Could you check is it good now or not? Should I also add information about added test in news?

rokm:
Great, thanks!

[Here](https://github.com/rokm/pyinstaller-hooks-contrib/runs/4337872835?check_suite_focus=true) is the oneshot test run with couple of last OpenCV releases.

The releases prior to introduction of the loader, 4.5.2.54 and 4.5.3.56, (still) work correctly and pass the test.
The 4.5.4.58 that introduced loader change is stil broken - it is hitting recursion error because our `FrozenImporter` does not work with module switching they are attempting to perform.
But what's really important here is that 4.5.4.60 is now fixed, and works out of the box. Which means the 4.5.4.58 and its issues can be forgotten, as users should be installing the latest available version.

That said, the changelog entry currently implies that 4.5.4.58 is also fixed, which it is not. Can we change that to 4.5.4.60, to make it clear that this is the version we support?

news/354.update.rst
```diff
@@ -0,0 +1,2 @@
+Update ``cv2`` hook to collect extra config files and modules for compatibility with OpenCV 4.5.4.58.
```
rokm, line 1: 4.5.4.58 -> 4.5.4.60, since 4.5.4.58 still does not work due to module switching attempt.

asenyaev: Sure, changed it to `4.5.4.60`. Just thought that it would work with `pyinstaller`, because python loader was added to OpenCV python package in `4.5.4.58` version.

rokm (approved):
Nice work, @asenyaev ! As far as I'm concerned, this is ready for merge. Let's see if @bwoodsend has any comments.

asenyaev:
@rokm, thank you for the mentoring! (:

bwoodsend:
Thanks @asenyaev!

> I'm sorry for such an unprepared pull request.

Pfff, we've seen far worse...

asenyaev:
@bwoodsend, could you please share the information when it will be released?

bwoodsend:
Done.

asenyaev:
Thank you a lot!

PR #356, opened 2021-11-25
Remove `twisted.internet.reactor` hook.

### LINKED ISSUE #355: Remove `twisted.internet.reactor` hook.
**Describe the bug**

As far as I can tell, the hook is no longer needed. In addition to that, having the hook installed means that installing a non-default reactor fails.

* Which hook/library isn't working? 
`twisted.internet.reactor`/`pyi_rth_twisted.py`

* Does the error get raised while building or when running?

Running.

**To Reproduce**

A minimal example file:
```python
from twisted.internet import asyncioreactor
asyncioreactor.install()
```

PyInstaller command:
```shell
pyinstall test.py
```

Error:
```
  File "test.py", line 2, in <module>
    asyncioreactor.install()
  File "twisted/internet/asyncioreactor.py", line 311, in install
  File "twisted/internet/main.py", line 31, in installReactor
twisted.internet.error.ReactorAlreadyInstalledError: reactor already installed
[...] Failed to execute script 'test' due to unhandled exception!
```

**Expected behavior**

I expect the above program to run without error.

 - OS: Fedora
 - Python Version: python 3.9.7
 ```
pyinstaller==4.7
pyinstaller-hooks-contrib==2021.3
```

**Additional context**

You can verify that things work without the plugin as described below. This is expected, as `twisted.internet.reactor` exists as a module that does exactly what the plugin does, after removing the imported path from `sys.modules`.

1. Create a do nothing hook:
    - Create an empty `rthooks/disabled.py`
    - Create `rthooks.dat` with
      ```
      { 'twisted.internet.reactor': ['disable.py'] }
      ```
2. Build the following script
    ```python
    from twisted.internet import reactor
    assert callable(reactor.listenTCP)
    ```
    with `pyinstaller test.py --additional-hooks-dir .`
3. Observer that running the built executable runs without errors.

### PR DESCRIPTION
**Pull request by tomprince**
Fix #355.

I also add tests to make sure both of the examples in the issue work.

### CODE REVIEW

File: src/_pyinstaller_hooks_contrib/tests/test_libraries.py [hunk 1]
```diff
@@ -995,3 +995,17 @@ def test_ffpyplayer(pyi_builder):
     pyi_builder.test_source("""
         import ffpyplayer.player
         """)
+
+
+@importorskip("twisted")
+def test_ffpyplayer(pyi_builder):
```
rokm, line 1001: The test name is probably a copy&paste typo.

Also, I think it would make sense to split this into two tests (e.g., `test_twisted_default_reactor` and `test_twisted_asyncio_reactor`) so each case is tested independently of the other.

tomprince: I've fixed both of those.

requirements-test-libraries.txt
```diff
@@ -66,9 +66,10 @@ git+https://github.com/shotgunsoftware/python-api.git@v3.2.6
 spacy==3.2.0
 srsly==2.4.2
 swagger-spec-validator==2.7.4
+tableauhyperapi==0.0.13821
```
bwoodsend, line 69: Why is this entry needed?

rokm: It was probably just moved to have the entries ordered alphabetically?

bwoodsend: Oh so it is. I didn't spot that it was a move rather than an addition.

bwoodsend:
[CI/CD run is here](https://github.com/bwoodsend/pyinstaller-hooks-contrib/actions/runs/1506726006). I don't suppose you know which version of twisted introduced the change that required this hook change? We're presumably breaking compatibility with some old version of twisted but it would be good to know how old.

rokm:
The runtime hook was introduced a long time ago via pyinstaller/pyinstaller@5dc89af8ebc545c9558fed7c4f4fc1e611438c9f, and references `twisted` 12.3.0 (Dec 26, 2012).

From what I can see, 17.9.0 (Sep 23, 2017) is the first version that supports python 3.6...

rokm:
For those failing CI/CD tests on Windows, [this](https://github.com/twisted/twisted/blob/a8aae6853b729a742ed4a99b95e8fe0c5f1dad97/src/twisted/internet/asyncioreactor.py#L35-L44) is probably relevant.

tomprince:
> I don't suppose you know which version of twisted introduced the change that required this hook change? We're presumably breaking compatibility with some old version of twisted but it would be good to know how old.

Testing with twisted 12.3.0, using python 2.7 and pyinstaller 3.6 patched to remove the twisted runtime hook, the first test included here passes. 

My guess is, is that if the patch was ever needed, it was changes in pyinstaller that removed the need for it, rather than anything on twisted's side. It seem plausible to me that `twisted/internet/reactor.py` mutation of `sys.modules` may have confused earlier versions pyinstaller, and that newer versions are more careful about managing imports of modules to be bundled.

bwoodsend:
OK, so anything since 12.3.0 doesn't need the hook but only versions >= 17.9 support any versions of Python that we support so we're not breaking anything.

Unfortunately now [you've knocked out python 3.6](https://github.com/bwoodsend/pyinstaller-hooks-contrib/actions/runs/1508526705)...

tomprince:
> OK, so anything since 12.3.0 doesn't need the hook but only versions >= 17.9 support any versions of Python that we support so we're not breaking anything.

I suspect even earlier versions work. :slightly_smiling_face: 

> Unfortunately now [you've knocked out python 3.6](https://github.com/bwoodsend/pyinstaller-hooks-contrib/actions/runs/1508526705)...

I added a condition to only set policy on py > 3.7.

rokm (approved):

bwoodsend:
Nice one @tomprince!

PR #366, opened 2022-01-09
Added pypsexec hook

arossert:


bwoodsend:
Can you give us a test too? Some instructions are [here](https://github.com/bwoodsend/pyinstaller-hooks-contrib/tree/readme#write-tests) if you need them.

arossert:
@bwoodsend I recently added another hook but I did not added a test for it, would you want me to?

rokm:
Aha, so the bundled .exe needs to be collected (as data file; on any platform) because it is deployed to the remote side. Can you add a comment to that effect to the hook itself, to make it clear what's going for anyone who might need to take a closer look at it at some later point? 

Otherwise, looks good!

bwoodsend:
Yeah, that's quite an interesting looking library.

@arossert Can you also [add pypsexec to the requirements file](https://github.com/bwoodsend/pyinstaller-hooks-contrib/tree/readme#pin-the-test-requirement) so that our CI/CD knows it needs to install it then you're good to go.

> @bwoodsend I recently added another hook but I did not add a test for it, would you want me to?

Yes please.

arossert:
@bwoodsend why do you fix the versions of the modules in requirements-test-libraries.txt? 
If there is a new release that breaks the hook it is a good place to find out about it.

bwoodsend:
There's a bot called PyUp which periodically bumps each package to its latest version so that we know if a new version breaks it (see any of the [PyUp bot's pull requests](https://github.com/pyinstaller/pyinstaller-hooks-contrib/pulls?q=is%3Apr+author%3Apyup-bot)). This kind of *rolling pin* allows us to keep a log of which older package versions worked.

bwoodsend:
You might want to set `git config --global user.email` to an email address registered with GitHub. Your commits aren't being properly linked to your account (although it doesn't matter here because I used squash-merge).

arossert:
@bwoodsend Thanks, the email I have in my global git setting is work-related, I changed it to my private email address on this repo 😃

PR #367, opened 2022-01-09
Added tests for mimesis

arossert:


src/_pyinstaller_hooks_contrib/tests/test_libraries.py
```diff
@@ -1048,3 +1048,11 @@ def test_pypsexec(pyi_builder):
         from pypsexec.paexec import paexec_out_stream
         next(paexec_out_stream())
         """)
+
+
+@importorskip("mimesis")
+def test_pypsexec(pyi_builder):
```
bwoodsend, line 1054:
```suggestion
def test_mimesis(pyi_builder):
```

requirements-test-libraries.txt
```diff
@@ -76,6 +76,7 @@ web3==5.25.0
 websockets==10.1; python_version >= '3.7'
 zeep==4.1.0
 pypsexec==0.3.0
+mimesis==5.2.1
```
bwoodsend, line 79: It seems that `mimesis` has dropped support for both Python 3.6 and 3.7 so this will need a python version constraint (like `websockets` has).

bwoodsend (approved):
And CI is happy too. https://github.com/bwoodsend/pyinstaller-hooks-contrib/actions/runs/1674974278

bwoodsend:
Thanks Amir!

PR #371, opened 2022-01-24
hooks: update tensorflow hook to support tensorflow 2.6 and later

rokm:
We need to explicitly collect keras package and its submodules due to lazy mapping of `tensorflow.keras.xyz` -> `keras.xyz`.

Fixes pyinstaller/pyinstaller#6519. Fixes pyinstaller/pyinstaller#6538.

### DIFF
--- a/news/371.update.rst
+++ b/news/371.update.rst
@@ -1,0 +1,1 @@
+Update ``tensorflow``  hook to add support for ``tensorflow`` 2.6.x and later.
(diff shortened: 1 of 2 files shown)


rokm:
Oneshot test: https://github.com/rokm/pyinstaller-hooks-contrib/actions/runs/1740278718 (the one failure is due to 2.8.0rc0 dropping support for python 3.7).

PR #377, opened 2022-02-07
Add hook for py to fix frozen pytest runner.

bwoodsend:
This fixes [the failing pytest-runner test](https://github.com/pyinstaller/pyinstaller-hooks-contrib/actions/runs/1803605477). [CI/CD](https://github.com/bwoodsend/pyinstaller-hooks-contrib/actions/runs/1808937412)

### DIFF
--- a/news/376.new.rst
+++ b/news/376.new.rst
@@ -1,0 +1,2 @@
+Add a hook for ``py`` which has dynamically loaded vendored submodules.
+This fixes compatibility with ``pytest >= 7.0.0``.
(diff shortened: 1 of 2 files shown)


rokm:
If I understand correctly, this is required by just-released `pytest 7.0.0`, which now performs the following in `_pytest.compat`:

```python
import py
LEGACY_PATH = py.path. local
```

But as this is not pytest-specific, should it be turned into a stand-alone test?

bwoodsend:
I suppose we could. The whole `py` package is pending deprecation so I doubt that it will evolve much and the `pytest-runner` test runs on every oneshot and PyUp triggered CI run (whether we want it to or not) so I thought that that would do.

rokm (approved):

PR #392, opened 2022-02-28
Hook for gitlab

albundy83:
https://github.com/albundy83/pyinstaller-hooks-contrib/actions/runs/1913214455


bwoodsend (approved):
I'm just doing a cross platform run for each of these. Other than that they're spot on.
https://github.com/bwoodsend/pyinstaller-hooks-contrib/actions/runs/1918032971

PR #396, opened 2022-03-15
Fix hook for imageio

paulmueller:
This update makes sure that all plugins from imageio are included. They are imported lazily and thus not found / ignored by PyInstaller.

https://imageio.readthedocs.io/en/stable/user_guide/freezing.html

https://github.com/imageio/imageio/issues/766


src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-imageio.py
```diff
@@ -12,6 +12,10 @@
...
 # Hook for imageio: http://imageio.github.io/
 
-from PyInstaller.utils.hooks import collect_data_files
+from PyInstaller.utils.hooks import collect_data_files, collect_submodules
 
 datas = collect_data_files('imageio', subdir="resources")
+
+# imageio plugins are imported lazily since version 0.11.0.
+# They are very light-weight, so we can safely include all of them.
+hiddenimports = collect_submodules('imageio.plugins')
```
FirefoxMetzger, line 21: I wonder how this interacts with transitive dependencies. Will this line only include the plugin (a single `.py` file inside `imageio/plugins`), or will this also add transitive dependencies/imports?

If it only adds the plugin file (shallow tracking) these are indeed lightweight, but this won't fully solve the problem, because backends may be left out. In this case I think it would be useful to document that users will have to explicitly add plugin-specific transitive dependencies (imageio-ffmpeg, tiffile, and consorts) as needed.

If the line does track and add transitive dependencies there are two points that come to mind: 

- Firstly, the plugin itself is leight-weight, but its transitive dependencies may not be so (thinking about ffmpeg, the freeImage library, etc. here). In this case it is possible to make a case for more selective importing depending on the use-case.
- Secondly, does pyinstaller gracefully handle missing modules (`ImportError` during module loading)? This is an implementation detail of ImageIO, but the way it currently works is that a plugin tries to load the backend (`import fancy_backend`) and if that fails, the import error is propagated to ImageIO core and properly resolved. If pyinsaller hooks into this process and attempts to discover transitive dependencies by importing plugin modules it, too, will see `ImportError`s for pluings with missing backends and will have to have a way to handle with this.

src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-imageio.py
```diff
@@ -12,6 +12,10 @@
 
 # Hook for imageio: http://imageio.github.io/
 
-from PyInstaller.utils.hooks import collect_data_files
+from PyInstaller.utils.hooks import collect_data_files, collect_submodules
 
 datas = collect_data_files('imageio', subdir="resources")
+
+# imageio plugins are imported lazily since version 0.11.0.
```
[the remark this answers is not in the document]
paulmueller: Yes, that was a typo.

src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-imageio.py
paulmueller: I just checked, according to https://github.com/pyinstaller/pyinstaller/blob/develop/PyInstaller/utils/hooks/__init__.py#L545, modules that cannot be imported will issue a warning and will then not be used. That means that if you don't have `imageio_ffmpeg` installed when you run PyInstaller, you will not have an ffmpeg binary in your distribution directory. I think this is fine and alignes well with how PyInstaller is meant to work.
But maybe someone from the PyInstaller developers would like to comment here as well.

rokm: If dependencies are present while freezing, they end up being pulled into the bundle. If not, they aren't, and the code trying to import the missing dependency will end up with a `ModuleNotFoundError`, as it would in unfrozen python. 

So someone using PyInstaller in a minimum build environment (i.e., with minimal requirements installed) ends up with a minimal build, while someone using a "universal" environment with lots of packages installed may end up with a bloated build. But as already noted above, that's how PyInstaller works.

As far as I can tell, this PR does not actually change the behavior w.r.t. collection of plugins' dependencies; the only change is that prior to 2.11.0, the plugins themselves were (all) automatically collected due to static `import`s in `imageio.plugins`, while from 2.11.0 they need to be collected using `collect_submodules` due to lazy imports. But in both cases, the transitive dependencies are collected if they are present.

PR #399, opened 2022-03-31
Add hook for Fiona hidden imports (dependencies of geopandas)

SMesser:
Tested with Fiona version 1.8.21 and PyInstaller 4.10.

See also [this SO question](https://stackoverflow.com/questions/56644056/pyinstaller-not-able-to-create-exe-using-geopandas-and-fiona-even-if-i-create-w/71681051#71681051)


SMesser:
> Hmmm, a basic `import fiona` example seems to also require `json` among hidden imports (probably imported by one of `fiona`'s compiled extensions).

That wasn't my experience, but... I'll try to update the `hiddenimports` based on your comment.  Not sure this will be right though.

SMesser:
and replicated your error... and my attempted fix here definitely won't be sufficient.  Unfortunately going to be a bit before I can push a patch.  Hopefully Saturday morning

  File "testfiona.py", line 1, in <module>
    import fiona
  File "PyInstaller/loader/pyimod03_importers.py", line 495, in exec_module
  File "fiona/__init__.py", line 86, in <module>
  File "PyInstaller/loader/pyimod03_importers.py", line 495, in exec_module
  File "fiona/collection.py", line 11, in <module>
  File "fiona/ogrext.pyx", line 6, in init fiona.ogrext
ModuleNotFoundError: No module named 'json'

src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-fiona.py
```diff
@@ -11,8 +11,7 @@
 # ------------------------------------------------------------------
 
 hiddenimports = [
-
     "fiona._shim",
-
+    "fiona.json",
```
rokm, line 15: Just `json` (the standard python module).

rokm:
> Hopefully Saturday morning

Sure, no rush :)

src/_pyinstaller_hooks_contrib/tests/test_libraries.py
```diff
@@ -16,6 +16,13 @@
 from PyInstaller.utils.tests import importorskip, requires, xfail
 
 
+def test_fiona(test):
```
bwoodsend, line 19: The argument name matters here. It has to be `pyi_builder` to pick up the custom `pyi_builder` [pytest fixture](https://docs.pytest.org/en/stable/explanation/fixtures.html). (It's not the most intuitive feature design.)

requirements-test-libraries.txt
```diff
@@ -19,6 +19,7 @@ dash==2.3.0
 dash-bootstrap-components==1.0.3
 dash-uploader==0.6.0
 fabric==2.7.0
+fiona==1.8.21
```
bwoodsend, line 22: Since fiona [doesn't readily support Windows](https://fiona.readthedocs.io/en/latest/README.html#windows). 
```suggestion
fiona==1.8.21; sys_platform != 'win32'
```

bwoodsend (changes requested):

bwoodsend (approved):
[CI passes](https://github.com/bwoodsend/pyinstaller-hooks-contrib/actions/runs/2084688396).

bwoodsend:
Nice work @SMesser

PR #400, opened 2022-03-31
Add hook for geopandas, which reads data files during import.

SMesser:
Tested for geopandas 0.10.2 and pyinstaller 4.10

See also these SO questions:
* https://stackoverflow.com/a/51957870
* https://stackoverflow.com/questions/56804095/pyinstaller-stopiteration-error-when-i-import-geopandas

### DIFF
--- a/news/400.new.rst
+++ b/news/400.new.rst
@@ -1,0 +1,1 @@
+Add geopandas data files for ``geopandas==0.10.2``.
--- a/requirements-test-libraries.txt
+++ b/requirements-test-libraries.txt
@@ -22,6 +22,7 @@ fabric==2.7.1
 fiona==1.8.21; sys_platform != 'win32'
 folium==0.12.1.post1
 ffpyplayer==4.3.5; python_version < '3.10' # doesn't have py310 wheels
+geopandas==0.10.2
 python-gitlab==3.9.0
 h5py==3.7.0; python_version >= '3.7'
 humanize==4.3.0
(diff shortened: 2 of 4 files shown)


rokm:
Can you please add a basic import test (which I think should be enough to reveal the need for those collected data files) [here](https://github.com/pyinstaller/pyinstaller-hooks-contrib/blob/master/src/_pyinstaller_hooks_contrib/tests/test_libraries.py), and add `geopandas` to the [list of packages to install during the CI test](https://github.com/pyinstaller/pyinstaller-hooks-contrib/blob/master/requirements-test-libraries.txt)?

Also, please add [a news entry for the PR](master/news/README.txt).

Thanks!

SMesser:
Fiona's a dependency of Geopandas, so #399 will probably have to merge first

madeline-scyphers:
Any updates on this? I have a project that needs geopandas and pyinstaller. Thanks!!!

bwoodsend:
Can we ignore Windows support for this like we do for Fionas? I don't fancy figuring out which goat we need to sacrifice to compile Fionas from source.

madeline-scyphers:
Yeah, that should probably be what you do. GDAL is such a pain in the but to get to install on windows if you are doing it with pip. fiona and geopandas both rely on GDAL, a C library. If you install geopandas/fiona with conda, you often have more success which is why geopandas recommends installing it with conda, it packages GDAL with the install. But since the CI job is using pip, ignoring windows is probably the easier solution.

madeline-scyphers:
It seems like geopandas in 0.11.0 removed python 3.7 support, but even on 0.10.2, some people had issues with it on 3.7 on mac os CI jobs because of things with pyproj I think. So maybe also bump it to 0.11.0 and limit it to python 3.8? Or drop it back to 0.10.0?

https://github.com/geopandas/geopandas/pull/2398
https://github.com/geopandas/geopandas/issues/2391

rokm (approved):

PR #406, opened 2022-04-11
Add a hook for clr_loader

rokm:
`clr_loader` is used by `pythonnet` 3.x, and on Windows, the default runtime (.NET framework) loader requires the package-bundled DLLs to be collected into the frozen application.

See: pyinstaller/pyinstaller#6572, pyinstaller/pyinstaller#6731.

### DIFF
--- a/news/406.new.rst
+++ b/news/406.new.rst
@@ -1,0 +1,3 @@
+Add a hook for ``clr_loader`` (used by upcoming ``pythonnet`` 3.x) that
+collects the DLLs required by the default runtime (.NET Framework) loader
+on Windows.
--- a/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-clr_loader.py
+++ b/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-clr_loader.py
@@ -1,0 +1,24 @@
+# ------------------------------------------------------------------
+# Copyright (c) 2022 PyInstaller Development Team.
+#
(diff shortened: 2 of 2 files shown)


rokm:
We already have a `test_pythonnet3`, but:
* it is inactive because on CI we install Windows-only `pythonnet` 2.x (and run `test_pythonnet2` instead)
* the test forces the .NET Core runtime, so it does not reveal the issue with default .NET Framework loader

We could set up a Windows-only test that only imports `clr`, but we'd need to ensure that .NET Framework is available on the CI runner (if it is not there by default...). But even that is assuming that we switch to git version of `pythonnet` until 3.x release gets out.

bwoodsend:
I'm not opposed to installing `pythonnet` from git. And doesn't [this](https://github.com/pyinstaller/pyinstaller-hooks-contrib/blob/7f740919a5533fb2c51a498e56dd1de04076d8ac/.github/workflows/oneshot-test.yml#L117) already set up the .NET framework?

rokm:
> I'm not opposed to installing `pythonnet` from git. And doesn't [this](https://github.com/pyinstaller/pyinstaller-hooks-contrib/blob/7f740919a5533fb2c51a498e56dd1de04076d8ac/.github/workflows/oneshot-test.yml#L117) already set up the .NET framework?

It sets up .NET Core 5.x, which is not the same as (Windows-only) .NET Framework (presumably 4.x).

bwoodsend (approved):

PR #408, opened 2022-04-11
hooks: update av hook for compatibility with >= 9.1.1

rokm:
From version 9.1.1 on, Windows PyPI wheels of av use a new DLL directory layout.


rokm:
Oneshot test: https://github.com/rokm/pyinstaller-hooks-contrib/actions/runs/2150926334

bwoodsend (approved):
Ooh, thanks for mopping that up. That's made my day.

PR #409, opened 2022-04-12
Add hook for lark

NathanVaughn:
I came across this issue when using [`commentjson`](https://pypi.org/project/commentjson/). Some Lark `.grammar` files that are needed were not being included when any sort of Lark `%import` was used. This should fix the issue for both the https://pypi.org/project/lark/ and older https://pypi.org/project/lark-parser/ packages (both install to `lark` inside `site-packages`).

Here are the CI results: https://github.com/NathanVaughn/pyinstaller-hooks-contrib/actions/runs/2157919610
Python 3.6 failed as it was unavailable, but I see that that the `setup.cfg` lists `>=3.7`.

I think I got everything, but let me know if I missed something!

### DIFF
--- a/news/409.new.rst
+++ b/news/409.new.rst
@@ -1,0 +1,1 @@
+Add a hook for ``lark`` (used by ``commentjson`` and others) that loads the needed grammar files.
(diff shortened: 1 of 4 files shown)


bwoodsend:
Nice one!

PR #420, opened 2022-04-28
Add hook for numcodecs

psavery:
numcodecs.compat_ext is only imported from pyx files, so it is missed.

Fixes: zarr-developers/numcodecs#291

### DIFF
--- a/news/420.new.rst
+++ b/news/420.new.rst
@@ -1,0 +1,1 @@
+Add a hook for ``numcodecs``, which has a hidden import.
--- a/requirements-test-libraries.txt
+++ b/requirements-test-libraries.txt
@@ -34,6 +34,7 @@ markdown==3.3.6
 # MetPy==1.2.0; python_version >= '3.7'
 mnemonic==0.20
 msoffcrypto-tool==5.0.0
+numcodecs==0.9.1; python_version < '3.10' # doesn't have python 3.10 wheels yet
 Office365-REST-Python-Client==2.3.11
 openpyxl==3.0.9
 pandas==1.4.1; python_version >= '3.8'
--- a/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-numcodecs.py
+++ b/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-numcodecs.py
@@ -1,0 +1,14 @@
+# ------------------------------------------------------------------
+# Copyright (c) 2022 PyInstaller Development Team.
+#
+# This file is distributed under the terms of the GNU General Public
+# License (version 2.0 or later).
+#
+# The full license is available in LICENSE.GPL.txt, distributed with
+# this software.
+#
+# SPDX-License-Identifier: GPL-2.0-or-later
+# ------------------------------------------------------------------
+
+# compat_ext is only imported from pyx files, so it is missed
+hiddenimports = ['numcodecs.compat_ext']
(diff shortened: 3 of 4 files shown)


bwoodsend:
Thanks for this. Can we get a [news entry](https://github.com/pyinstaller/pyinstaller-hooks-contrib#add-a-news-entry) and preferably also [a test](https://github.com/pyinstaller/pyinstaller-hooks-contrib#test).

psavery:
> Thanks for this. Can we get a [news entry](https://github.com/pyinstaller/pyinstaller-hooks-contrib#add-a-news-entry) and preferably also [a test](https://github.com/pyinstaller/pyinstaller-hooks-contrib#test)?

News entry and test are added!

bwoodsend:
Since you are adding a test - can you also add:
```
numcodes==0.9.1; python_version < '3.10' # doesn't have python 3.10 wheels yet
```
to `requirements-test-libraries.txt`? Then we're done.

psavery:
> Since you are adding a test - can you also add:
> 
> ```
> numcodes==0.9.1; python_version < '3.10' # doesn't have python 3.10 wheels yet
> ```
> 
> to `requirements-test-libraries.txt`? Then we're done.

Done

jakirkham:
JFYI we do include Python 3.10 in the 0.10.0 pre-releases (alphas). So this can be relaxed in the future when 0.10.0 is used here.

bwoodsend:
Hmmm, it appears to be stalling indefinitely on macOS Python 3.8 and 3.9. Do you have a macOS to test on? If you don't, I'll have a look at it tomorrow.

rokm:
I think the problem is that `numcodecs` internally uses `multiprocessing` (e.g., [here](https://github.com/zarr-developers/numcodecs/blob/15f17138b9b41e99faa4a2670e2580a782547d6b/numcodecs/__init__.py#L47)), so without `multiprocessing.freeze_support()` at the start of program/test, we end up with infinite spawn loop ([before](https://github.com/rokm/pyinstaller-hooks-contrib/runs/6225567330?check_suite_focus=true) and [after](https://github.com/rokm/pyinstaller-hooks-contrib/runs/6225880362?check_suite_focus=true)).

---

On an unrelated note, I wonder if there's any limit on how early in the code `freeze_support()` needs to be called - i.e., would it work if we simply called it in the runtime hook, after we define and override the original version (we could perhaps even skip the override part)? 

It would go a long way if we could alleviate the need to explicitly call `freeze_support` somehow, especially in cases like this where `multiprocessing` use is buried in some 3rd party module...

psavery:
> I think the problem is that `numcodecs` internally uses `multiprocessing` (e.g., [here](https://github.com/zarr-developers/numcodecs/blob/15f17138b9b41e99faa4a2670e2580a782547d6b/numcodecs/__init__.py#L47)), so without `multiprocessing.freeze_support()` at the start of program/test, we end up with infinite spawn loop ([before](https://github.com/rokm/pyinstaller-hooks-contrib/runs/6225567330?check_suite_focus=true) and [after](https://github.com/rokm/pyinstaller-hooks-contrib/runs/6225880362?check_suite_focus=true)).

I cherry-picked and pushed your fix.

psavery:
A side note: for macOS, the spawn multiprocessing start method [is the default as of python 3.8](https://docs.python.org/3/library/multiprocessing.html#contexts-and-start-methods), which is likely why the tests passed for python 3.7.

bwoodsend:
> On an unrelated note, I wonder if there's any limit on how early in the code freeze_support() needs to be called - i.e., would it work if we simply called it in the runtime hook, after we define and override the original version (we could perhaps even skip the override part)?

Yeah, it's been on the edge of my mind to figure out why we can't just call `freeze_support()` at the beginning as soon as a frozen application process has figured out that it's not a subprocess but I never looked at it properly.

bwoodsend:
Thanks @psavery

PR #436, opened 2022-05-25
hooks: pylint: fix filter function passed to collect_submodules

### LINKED ISSUE #435: `hook-pylint.py` filter seems to be out of date
https://github.com/pyinstaller/pyinstaller-hooks-contrib/blob/8f66a66b89e76e6213972fde1c84a337b11fbd5d/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-pylint.py#L63-L66

Line 65: [I don't think the `pylint.test` package exists (at least not currently)](https://github.com/PyCQA/pylint/tree/main/pylint).

Line 66:
```python
not name == 'testutils'
```
should really be
```python
not is_module_or_submodule(name, 'pylint.testutils')
```

**Desktop:**
 - OS: Windows 10
 - Python Version: Python 3.9.5
 - Version of `pyinstaller-hooks-contrib`: 2022.4
 - Version of PyInstaller: 5.1

**Additional context**
Mentioned in https://github.com/pyinstaller/pyinstaller/issues/6846

rokm:
Fix filter function passed to `collect_submodules` to correctly exclude both `pylint.test` sub-package (available in `pylint` <= 2.3) and `pylint.testutils submodule` (`pylint` < 2.7) or subpackage (>= `2.7`).

Previously, `pylint.testutils` and its contents were not excluded.

Closes #435.

### DIFF
--- a/news/435.update.rst
+++ b/news/435.update.rst
@@ -1,0 +1,2 @@
+Fix the filter function used with ``collect_submodules`` in the ``pylint``
+hook to properly exclude ``pylint.testutils``.
--- a/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-pylint.py
+++ b/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-pylint.py
@@ -60,7 +60,15 @@ datas = (
          collect_data_files('pylint.reporters', True)
          )
 
-# Add imports from dynamically loaded modules excluding tests and testutils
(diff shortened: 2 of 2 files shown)


rokm:
Once we fix `collect_submodules` to avoid recursing into filtered-out sub-packages, the filter condition could be simplified to `name not in {'pylint.test', 'pylint.testutils'}`. But the current approach is backwards-compatible.

PR #437, opened 2022-05-26
Add hook-pypemicro.py

cz6ace:
Tests included

### DIFF
--- a/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-pypemicro.py
+++ b/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-pypemicro.py
@@ -13,7 +13,7 @@
 # Hook for the pypemicro module: https://github.com/nxpmicro/pypemicro
 
 import os
-from PyInstaller.utils.hooks import remove_prefix, get_package_paths, is_module_satisfies
+from PyInstaller.utils.hooks import get_package_paths, is_module_satisfies
 from PyInstaller.log import logger
 from PyInstaller.compat import is_darwin
 
@@ -31,13 +31,13 @@ if is_module_satisfies('pyinstaller >= 5.0'):
     for lib in get_safe_libs():
         source_path = lib['path']
         source_name = lib['name']
-        dest = os.path.relpath(source_path, os.path.dirname(pkg_base))
+        dest = os.path.relpath(source_path, pkg_base)
         binaries.append((os.path.join(source_path, source_name), dest))
         if is_darwin:
             libusb = os.path.join(source_path, 'libusb.dylib')
             if os.path.exists(libusb):
                 binaries.append((libusb, dest))
             else:
-                logger.warning(f"libusb.dylib was not found for Mac OS, ignored")
+                logger.warning("libusb.dylib was not found for Mac OS, ignored")
 else:
     logger.warning("hook-pypemicro requires pyinstaller >= 5.0")
--- a/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-pypemicro.py
+++ b/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-pypemicro.py
@@ -1,0 +1,43 @@
+# ------------------------------------------------------------------
+# Copyright (c) 2022 PyInstaller Development Team.
+#
+# This file is distributed under the terms of the GNU General Public
+# License (version 2.0 or later).
+#
+# The full license is available in LICENSE.GPL.txt, distributed with
+# this software.
+#
+# SPDX-License-Identifier: GPL-2.0-or-later
+# ------------------------------------------------------------------
+
+# Hook for the pypemicro module: https://github.com/nxpmicro/pypemicro
+
+import os
+from PyInstaller.utils.hooks import remove_prefix, get_package_paths, is_module_satisfies
+from PyInstaller.log import logger
+from PyInstaller.compat import is_darwin
+
+binaries = list()
+if is_module_satisfies('pyinstaller >= 5.0'):
+    from PyInstaller import isolated
+
+    @isolated.decorate
+    def get_safe_libs():
+        from pypemicro import PyPemicro
+        libs = PyPemicro.get_pemicro_lib_list()
+        return libs
+
+    pkg_base, pkg_dir = get_package_paths("pypemicro")
+    for lib in get_safe_libs():
+        source_path = lib['path']
+        source_name = lib['name']
+        dest = os.path.relpath(source_path, os.path.dirname(pkg_base))
+        binaries.append((os.path.join(source_path, source_name), dest))
+        if is_darwin:
+            libusb = os.path.join(source_path, 'libusb.dylib')
+            if os.path.exists(libusb):
+                binaries.append((libusb, dest))
+            else:
+                logger.warning(f"libusb.dylib was not found for Mac OS, ignored")
+else:
+    logger.warning("hook-pypemicro requires pyinstaller >= 5.0")
--- a/news/417.new.rst
+++ b/news/417.new.rst
@@ -1,0 +1,1 @@
+Add libraries for ``pypemicro==0.1.9``
--- a/requirements-test-libraries.txt
+++ b/requirements-test-libraries.txt
@@ -52,6 +52,7 @@ pycryptodome==3.14.1
 pycryptodomex==3.14.1
 pyexcelerate==0.10.0
 pylint==2.13.9
+pypemicro==0.1.9
 pyphen==0.12.0
 pyppeteer==1.0.2
 pyusb==1.2.1
--- a/src/_pyinstaller_hooks_contrib/tests/test_libraries.py
+++ b/src/_pyinstaller_hooks_contrib/tests/test_libraries.py
@@ -1144,6 +1144,14 @@ def test_numcodecs(pyi_builder):
     """)
 
 
+@importorskip('pypemicro')
+def test_pypemicro(pyi_builder):
+    pyi_builder.test_source("""
+        from pypemicro import PyPemicro
+        assert PyPemicro.get_pemicro_lib()
+    """)
+
+
 @importorskip('sounddevice')
 def test_sounddevice(pyi_builder):
     pyi_builder.test_source("""


src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-pypemicro.py
```diff
@@ -0,0 +1,75 @@
...
+                if destdir:
+                    # Libraries will be put in the same directory.
+                    dest = destdir
+                else:
+                    # The directory hierarchy is preserved as in the original package.
+                    dest = remove_prefix(dirpath, os.path.dirname(pkg_base) + os.sep)
+                logger.debug(' %s, %s' % (source, dest))
+                dylibs.append((source, dest))
+    return dylibs
+
```
bwoodsend, line 62: What's the motivation for copy/pasting this function in here instead of just using `from PyInstaller.utils.hooks import collect_dynamic_libs`?

cz6ace: I forgot to add a comment. The important change is in `PY_DYLIB_PATTERNS` where `lib*.so` has been changed into `*.so`
So instead of writing new (buggy) code I copied the existing just with this change.

cz6ace: Yes, I wanted to avoid importing the lib. The isolated subprocess is new to me. I'll give it a try, thanks for pointing that.

cz6ace: Refactored using the isolation and calling the pypemicro for libraries.

src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-pypemicro.py
```diff
@@ -0,0 +1,35 @@
...
+    libs = PyPemicro.get_pemicro_lib_list()
+    return libs
+
+
+pkg_base, pkg_dir = get_package_paths("pypemicro")
+
+binaries = list()
+for lib in get_safe_libs():
+    source = lib['path']
+    for file in glob.glob(os.path.join(source, '*')):
```
rokm, line 33: Why do we need to glob the library directory? Shouldn't we just be collecting the `os.path.join(lib['path'], lib['name'])` instead? Otherwise, we'll be collecting both 32-bit and 64-bit library wheras only one of them is really required. 

I see that macOS also ships `libusb.dylib`, but I see no indication that it is used anywhere (it is not link-time dependency of `unitacmp-64.dylib` nor is it loaded dynamically by the python code. Might be dynamically loaded by some function in `unitacmp-64.dylib`, though?)

src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-pypemicro.py
```diff
@@ -0,0 +1,35 @@
...
+        dest = remove_prefix(source, os.path.dirname(pkg_base) + os.sep)
```
rokm, line 34: Use `os.path.relpath` instead of `remove_prefix`. That way you don't need to worry about trailing separators.

src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-pypemicro.py
```diff
@@ -0,0 +1,35 @@
...
+
+# Hook for the pypemicro module: https://github.com/nxpmicro/pypemicro
+
+import glob
+import os
+from PyInstaller import isolated
+from PyInstaller.utils.hooks import remove_prefix, get_package_paths
+
+
+@isolated.decorate
```
rokm, line 21: Since `isolated` is available only in PyInstaller 5, we need to decide what to do about earlier versions. Either we'll need to tie the next release of contrib hooks to PyInstaller 5.0 and later, or this hook will need to check for PyInstaller version and become no-op (or raise error) for older versions (assuming we won't be implementing alternative codepath with old-style subprocesses - which I'd prefer to avoid). @bwoodsend, what do you think?

src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-pypemicro.py
rokm: > I see that macOS also ships `libusb.dylib`, but I see no indication that it is used anywhere (it is not link-time dependency of `unitacmp-64.dylib` nor is it loaded dynamically by the python code. Might be dynamically loaded by some function in `unitacmp-64.dylib`, though?)

Hmm, the strings in `unitacmp-64.dylib` seem to imply that `libusb.dylib` might be dynamically loaded somewhere. Still, I think it might be better to collect just `os.path.join(lib['path'], lib['name'])` on all OSes, and in addition collect `os.path.join(lib['path'], 'libusb.dylib')` on macOS.

src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-pypemicro.py
bwoodsend: Uhm, yeah I was just going to wing it and leave it unchecked. This repo doesn't enforce any minimum PyInstaller version (it doesn't even have PyInstaller as a dependency) and TBH I'm not convinced that it should force users onto newer PyInstaller versions because some random hook that they're likely not using requires some new feature. I wouldn't mind adding a little check at the top of the hook with a meaningful error message.

cz6ace: well, do you mean throw an exception for pyinstaller < 5.0 or just ERROR message using logger?

rokm: Compatibility-wise, I think the best approach would be to put the contents of the hook under the PyInstaller version check; if pyinstaller < 5.0, display a warning and do nothing else.

rokm: Or even put the `isolated` code part under the version check and return empty library list in the else branch, along with the warning. This way, interested parties can add a fallback using old `exec_statement`...

src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-pypemicro.py
```diff
@@ -0,0 +1,39 @@
...
+
+# Hook for the pypemicro module: https://github.com/nxpmicro/pypemicro
+
+import os
+from PyInstaller.utils.hooks import remove_prefix, get_package_paths, is_module_satisfies
+from PyInstaller.log import logger
+from PyInstaller.compat import is_darwin
+
+binaries = list()
+if is_module_satisfies('pyinstaller < 5.0'):
```
rokm, line 21: This is probably a matter of personal preference, but I'd rather see `is_module_satisfies('pyinstaller >= 5.0')` here (and consequently warning in the else branch).

src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-pypemicro.py
cz6ace: @rokm, glob removed, Mac OS handled explicitly. check + warning for pyinstaller < 5.0 implemented

src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-pypemicro.py
```diff
@@ -0,0 +1,39 @@
...
+        return libs
+
+    pkg_base, pkg_dir = get_package_paths("pypemicro")
+    for lib in get_safe_libs():
+        source_path = lib['path']
+        source_name = lib['name']
+        dest = remove_prefix(source_path, os.path.dirname(pkg_base) + os.sep)
+        binaries.append((os.path.join(source_path, source_name), dest))
+        if is_darwin:
+            binaries.append((os.path.join(source_path, 'libusb.dylib'), dest))
```
rokm, line 39: We should probably check that `os.path.join(source_path, 'libusb.dylib')` exists before trying to add it.

rokm (commented):
Nicely done. Couple more nitpicks, then it should be ready for merge.

src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-pypemicro.py
```diff
@@ -0,0 +1,39 @@
...
+    def get_safe_libs():
+        from pypemicro import PyPemicro
+        libs = PyPemicro.get_pemicro_lib_list()
+        return libs
+
+    pkg_base, pkg_dir = get_package_paths("pypemicro")
+    for lib in get_safe_libs():
+        source_path = lib['path']
+        source_name = lib['name']
+        dest = remove_prefix(source_path, os.path.dirname(pkg_base) + os.sep)
```
rokm, line 36: Ah, and I'd still prefer to see `remove_prefix` replaced with `os.path.relpath`.

src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-pypemicro.py
cz6ace: done

src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-pypemicro.py
cz6ace: Oh, I missed this one, will do another (squashed) commit

cz6ace:
@rokm , please check. Now or never ;-)

bwoodsend (approved):

src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-pypemicro.py
```diff
@@ -0,0 +1,43 @@
...
+    def get_safe_libs():
+        from pypemicro import PyPemicro
+        libs = PyPemicro.get_pemicro_lib_list()
+        return libs
+
+    pkg_base, pkg_dir = get_package_paths("pypemicro")
+    for lib in get_safe_libs():
+        source_path = lib['path']
+        source_name = lib['name']
+        dest = os.path.relpath(source_path, os.path.dirname(pkg_base))
```
rokm, line 34: I think this should be using `pkg_base` instead of `os.path.dirname(pkg_base))`. 

Both work as expected, but that's only because `pkg_base`  has a trailing separator, and so `os.path.dirname(pkg_base)` still returns the path to the site-packages directory. So while it works, it gives incorrect impression that we are after the parent directory, which we are not..

cz6ace: you are right, it is incorrect, probably leftover from the `remove_prefix`

rokm (approved):

PR #410, opened 2022-04-13
update hook-weasyprint.py

seppovic:
weasyprint tries to `ffi.dlopen()` some GTK+ libraries. With the update, these are added to binaries array at least on windows (I don't know how to find the files on the various linux distributions)

See https://doc.courtbouillon.org/weasyprint/stable/first_steps.html#windows and https://github.com/Kozea/WeasyPrint/blob/master/weasyprint/text/ffi.py#L388-L413 for reference.

PR #439, opened 2022-05-27
Update Weasyprint-hook

wf-r:
This PR updates the Weasyprint hook to work with Weasyprint 54.0 and later.

In contrast to https://github.com/pyinstaller/pyinstaller-hooks-contrib/pull/410 it should work on all OS (by using the established calls to load libraries).


wf-r:
Failing checks was an issue of Github not responding.

wf-r:
Could anybody test this on Linux or MacOS and ensure that importing Weasyprint does not log a `Fontconfig error: Cannot load default config file: No such file: (null)` warning during execution (test cases in the code). On Windows it does work, however I fear that the /etc/fonts directory needs separate handling on other OS.

bwoodsend:
Sorry, I should have commented on #410. Bundling gi/Gtk is such a messy business that it's got its own relatively high maintenance corners of PyInstaller (see [here](https://github.com/pyinstaller/pyinstaller/blob/15f23a8a89be5453b3520df8fc3e667346e103a6/PyInstaller/utils/hooks/gi.py) and all the `hook-gi*.py` files in [here](https://github.com/pyinstaller/pyinstaller/tree/790584f7917ea0398d55af7a61f381854f09c23f/PyInstaller/hooks)). So I'm pretty reluctant to accept anything which depends on Gtk, in particular a DIY implementation that does its own Gtk discovery rather than using `PyInstaller.utils.hook.gi` and in particular one that's untestable on CI/CD (although [they just about manage it](https://github.com/Kozea/WeasyPrint/blob/26d5d8264d33168120ffaa93218ba013b1a5c784/.github/workflows/tests.yml#L22-L39) so it probably is doable).

rokm:
The problem is, Gtk and associated libraries can be used without GObjectIntrospection (gi), as is the case here. So the hook is correct to bypass all the `gi` stuff, because `gi` might not even be installed.

I'd really prefer if we could avoid searching for the library patterns in the hook and instead look up for example `weasyprint.text.ffi.pango`, but unfortunately, I don't see a way to get the loaded library path from a `cffi.api._make_ffi_library.<locals>.FFILibrary` object... So perhaps the search as it is currently implemented is the best we can do (and it is already an improvement over #410).

As for the tests, I suspect that on linux (and possibly also under homebrew on macOS), the collected fontconfig still looks at the system /etc/fonts, so the tests passing on CI do not mean much.

bwoodsend:
Fine then. @wf-r Give this a rebase to deal with the merge conflict then I'll merge this.

wf-r:
@rokm @bwoodsend Thanks for the insights. I have merged the latest changes, and added a statement to try to load `/etc/fonts` if available.

bwoodsend (approved):
I'll leave @rokm to hit the merge button.

rokm:
Thanks @wf-r!

PR #444, opened 2022-06-09
hooks: update clr hook for PyInstaller 5.1 compatibility

rokm:
Add `platform` and `warnings` to hidden imports; previously, these were collected as part of `pickle` -> `doctest` -> ... optional import chain, which was removed in PyInstaller/PyInstaller#6798.

Fixes PyInstaller/PyInstaller#6882.


rokm:
[Here](https://github.com/rokm/pyinstaller-hooks-contrib/actions/runs/2467259902) is a "before" oneshot test, which reproduces https://github.com/pyinstaller/pyinstaller/issues/6882.

bwoodsend (approved):

PR #449, opened 2022-06-14
Add hook pyshark

LpCodes:
Have made changes as suggest kindly review


rokm (changes requested):
See the inline comments. Also, make your changes to this PR instead of opening a new one. Thanks!

src/_pyinstaller_hooks_contrib/tests/test_libraries.py
```diff
@@ -1196,3 +1196,14 @@ def test_great_expectations(pyi_builder):
         from great_expectations.render.view import view
         v = view.DefaultJinjaView()
         """)
+
```
rokm, line 1199: There must be two empty lines between individual tests (this is part of why linter check is failing).

rokm (changes requested):

src/_pyinstaller_hooks_contrib/tests/test_libraries.py
LpCodes: Added 2 empty lines

bwoodsend:
Thanks @LpCodes!

PR #451, opened 2022-06-15
hooks: add pre-safe-import hook for tensorflow >= 2.8.0

rokm:
Register the alias `tensorflow.keras` = `keras.api._v2.keras`, which is required as of `tensorflow >= 2.8.0` for programs that import
only from `tensorflow.keras`.

Add a new import test for `tensorflow.keras` that demonstrates the issue. 

See https://github.com/pyinstaller/pyinstaller/discussions/6890

### DIFF
--- a/requirements-test-libraries.txt
+++ b/requirements-test-libraries.txt
@@ -100,6 +100,7 @@ soundfile==0.10.3.post1; sys_platform != 'linux'
 yt-dlp==2022.5.18
 limits==2.6.3
 great-expectations==0.15.9
+tensorflow==2.9.1
 
 # ------------------- Platform (OS) specifics
 
--- a/news/451.update.rst
+++ b/news/451.update.rst
@@ -1,0 +1,2 @@
+Fix ``tensorflow`` not being collected at all when using ``tensorflow``
+2.8.0 or newer and importing only from the ``tensorflow.keras`` subpackage.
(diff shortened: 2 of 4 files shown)


rokm:
[Before](https://github.com/rokm/pyinstaller-hooks-contrib/actions/runs/2498173317) (but with added new test) and [after](https://github.com/rokm/pyinstaller-hooks-contrib/actions/runs/2504937601).

bwoodsend:
Shall we add tensorflow to `requirements-test-libraries.txt`?

bwoodsend (approved):

rokm:
Hmm, I suppose with the incremental CI regime that we have now, we could do that...

rokm:
Not that I'm looking forward to being informed what the latest `tensorflow` release broke in PyInstaller, but I suppose that's going to happen one way or another.

requirements-test-libraries.txt
```diff
@@ -100,6 +100,7 @@ soundfile==0.10.3.post1; sys_platform != 'linux'
 yt-dlp==2022.5.18
 limits==2.6.3
 great-expectations==0.15.9
+tensorflow==2.9.1
```
bwoodsend, line 103: That's an ironic place to put tensorflow - anywhere near the words *great expectations*.

bwoodsend: We need to find a package named *blood-sweat-and-carnage* to stick it next to instead.

rokm: > We need to find a package named _blood-sweat-and-carnage_ to stick it next to instead.

That sounds like an alias for PyInstaller.

bwoodsend (approved):

PR #453, opened 2022-06-27
 hooks: cv2: collect Qt fonts data and plugins from linux PyPI wheels

rokm:
In linux PyPI `opencv-python` wheels, the `cv2` extension is linked against Qt, and the wheel bundles a basic subset of Qt shared libraries, plugins, and font files. This is not the case on other OSes (presumably native UI APIs are used by the OpenCV HighGUI
module), nor in the headless PyPI wheels (`opencv-python-headless`).

The bundled Qt shared libraries should be picked up automatically due to binary dependency analysis, but we need to collect  plugins and font files from the `qt` subdirectory.

### DIFF
--- a/news/453.update.rst
+++ b/news/453.update.rst
@@ -1,0 +1,2 @@
+(Linux) Ensure that OpenCV hook collects Qt plugins and font files that
+are bundled with linux versions of ``opencv-python`` PyPI wheels.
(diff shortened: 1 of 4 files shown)


bwoodsend (approved):
Do we not need a news entry for this?

rokm:
> Do we not need a news entry for this?

We do.

PR #455, opened 2022-06-27
hooks: add hook for pandas_flavor

rokm:
Add hook for `pandas_flavor` to handle hidden imports caused by switch to lazy loader in version 0.3.0 of the package.


bwoodsend (approved):
Damn you for being so proactive!

PR #460, opened 2022-07-05
add zstandard libraries from hdf5plugin for dclab hook

paulmueller:
The package dclab uses the Zstandard compression algorithm HDF5 plugin from the hdf5plugin package. This pull request makes sure that the corresponding dll/so/dylib files are present in the distribution. I am the author of dclab.
https://github.com/ZellMechanik-Dresden/dclab

### DIFF
--- a/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-dclab.py
+++ b/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-dclab.py
@@ -15,3 +15,6 @@
 from PyInstaller.utils.hooks import collect_data_files
 
 datas = collect_data_files('dclab')
+
+# Add the Zstandard library used by dclab by default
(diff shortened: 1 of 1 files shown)


bwoodsend:
Any reason why this should go in the hook for `dclab` instead of a hook for `hdf5plugin`?

paulmueller:
My initial idea was this: If I used a general hook for hdf5plugin, then **all** of the library files (not only Zstandard) would be added. I intended to only add Zstandard here, because none of the other libraries are required.

But I see that a hook for hdf5plugin might be a better, more general solution, even if it means that it increases the distribution size.

Should I create an hdf5plugin hook instead?

bwoodsend:
If hdf5plugin is usable without those extra files then no.

paulmueller:
Then that means yes, because in the most general case, you will need those extra files with hdf5plugin. I will create a hook for hdf5plugin.

PR #463, opened 2022-07-08
ci: install MariaDB connector from official MariaDB CS repository

rokm:
Install MariaDB Connector/C from official MariaDB Community Server repository. The version shipped with Ubuntu 20.04 is too old and is not supported by `mariadb` >= 1.1.3.

### DIFF
--- a/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-mariadb.py
+++ b/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-mariadb.py
@@ -10,9 +10,17 @@
 # SPDX-License-Identifier: GPL-2.0-or-later
 # ------------------------------------------------------------------
 
+from PyInstaller.utils.hooks import is_module_satisfies, collect_submodules
+
 # The MariaDB uses a .pyd file that imports ``decimal`` module within its
 # module initialization function. On recent python versions (> 3.8), the decimal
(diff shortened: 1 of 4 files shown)


src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-mariadb.py
```diff
@@ -10,9 +10,17 @@
...
+
 # The MariaDB uses a .pyd file that imports ``decimal`` module within its
 # module initialization function. On recent python versions (> 3.8), the decimal
 # module seems to be picked up nevertheless (presumably due to import in some
 # other module), but it is better not to rely on that, and ensure it is always
 # collected as a hidden import.
 hiddenimports = ['decimal']
+
+
+# mariadb >= 1.1.0 requires several of hidden imports from mariadb.constants.
```
bwoodsend, line 23:
```suggestion
# mariadb >= 1.1.0 requires several hidden imports from mariadb.constants.
```

bwoodsend (approved):

src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-mariadb.py
rokm: Good catch! That's what happens when "couple of" gets changed into "several", and you don't pay attention...

PR #466, opened 2022-07-11
Fix hook-pyqtgraph for PyInstaller 5.2

### LINKED ISSUE #465: pyqtgraph hook no longer finds hidden imports since PyInstaller 5.2
Hi,

When an application that uses `pyqtgraph` is frozen with PyInstaller 5.2, at runtime we get errors like:

```
  File "pyqtgraph\graphicsItems\ViewBox\ViewBoxMenu.py", line 6, in <module>
  File "importlib\__init__.py", line 126, in import_module
ModuleNotFoundError: No module named 'pyqtgraph.graphicsItems.ViewBox.axisCtrlTemplate_pyqt5' 
```

The `hook-pyqtgraph` file contains code to dynamically load some files:

https://github.com/pyinstaller/pyinstaller-hooks-contrib/blob/a3ddd1d597834733e5555e3537cdd5080958ee85/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-pyqtgraph.py#L19-L35

I added some debugging code to it and noticed that `hiddenimports` is `[]`, so digging the code history I could trace the problem to this commit:

https://github.com/pyinstaller/pyinstaller/commit/17aa452741f71b9ac4bab0b3cbe566d68a0bdf5f

I will open a PR to fix this shortly. 👍

nicoddemus:
In 5.2 the collect_submodules function no longer recurses into
filtered packages (introduced in https://github.com/pyinstaller/pyinstaller/commit/17aa452741f71b9ac4bab0b3cbe566d68a0bdf5f), which broke the pyqtgraph hooks which relied on that behavior.

This fixes the hook by collecting all modules with `collect_submodules`, and filtering
for the hidden modules manually.

Took the opportunity to add tests for `pyqtgraph`.

Fixes #465

### DIFF
--- a/src/_pyinstaller_hooks_contrib/tests/test_libraries.py
+++ b/src/_pyinstaller_hooks_contrib/tests/test_libraries.py
@@ -1238,6 +1238,7 @@ def test_pyshark(pyi_builder):
 
 
 @importorskip('pyqtgraph')
+@importorskip('PyQt5')
 def test_pyqtgraph(pyi_builder):
     pyi_builder.test_source(
         """
--- a/src/_pyinstaller_hooks_contrib/tests/test_libraries.py
+++ b/src/_pyinstaller_hooks_contrib/tests/test_libraries.py
@@ -1235,3 +1235,14 @@ def test_pyshark(pyi_builder):
         #print(data)
         """
     )
+
+
+@importorskip('pyqtgraph')
+def test_pyqtgraph(pyi_builder):
+    pyi_builder.test_source(
+        """
+        import pyqtgraph.graphicsItems.PlotItem
+        import pyqtgraph.graphicsItems.ViewBox.ViewBoxMenu
+        import pyqtgraph.imageview.ImageView
+        """
+    )
--- a/.github/workflows/pr-test.yml
+++ b/.github/workflows/pr-test.yml
@@ -37,6 +37,7 @@ jobs:
           > requirements-test-libraries.txt
           set -e
           echo '-r requirements-test.txt' >> requirements-test-libraries.txt
+          if grep -q pyqtgraph requirements-test-libraries.txt ; then echo PyQt5 >> requirements-test-libraries.txt ;fi
           cat requirements-test-libraries.txt
 
       - name: Set up .NET Core for pythonnet tests
--- a/news/465.bugfix.rst
+++ b/news/465.bugfix.rst
@@ -1,0 +1,1 @@
+Fixed ``pyqtgraph`` hook for PyInstaller 5.2.
--- a/requirements-test-libraries.txt
+++ b/requirements-test-libraries.txt
@@ -55,6 +55,7 @@ pylint==2.14.4
 pypemicro==0.1.9
 pyphen==0.12.0
 pyppeteer==1.0.2
+pyqtgraph==0.12.4
 pyusb==1.2.1
 pyviz-comms==2.2.0
 pyvjoy==1.0.1; sys_platform == 'win32'
--- a/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-pyqtgraph.py
+++ b/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-pyqtgraph.py
@@ -28,8 +28,9 @@ datas = collect_data_files("pyqtgraph",
 # - pyqtgraph.graphicsItems.PlotItem.plotConfigTemplate_pyside2
 # - pyqtgraph.graphicsItems.PlotItem.plotConfigTemplate_pyside6
 #
-# To be future-proof, we include all of them via a filter in
-# collect-submodules.
+# To be future-proof, we collect all modules by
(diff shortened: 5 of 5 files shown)


src/_pyinstaller_hooks_contrib/tests/test_libraries.py
```diff
@@ -1235,3 +1235,14 @@ def test_pyshark(pyi_builder):
         #print(data)
         """
     )
+
+
+@importorskip('pyqtgraph')
+def test_pyqtgraph(pyi_builder):
+    pyi_builder.test_source(
+        """
+        import pyqtgraph.graphicsItems.PlotItem
```
nicoddemus, line 1244: Those modules have dynamic imports for the template modules that the hook declares as hidden. 👍

nicoddemus:
The tests failed because no PyQt library could be imported, but I see `PyQt5` listed in `requirements-test-libraries.txt`, any hints?

bwoodsend:
> The tests failed because no PyQt library could be imported, but I see PyQt5 listed in requirements-test-libraries.txt, do I have to do something else for PyQt5 to be installed?

[The requirements file is filtered](https://github.com/pyinstaller/pyinstaller-hooks-contrib/blob/a3ddd1d597834733e5555e3537cdd5080958ee85/.github/workflows/pr-test.yml#L28-L40) so that only entries changed are installed. Otherwise our CI takes days testing hooks that aren't touched by the PR (actually now it fails almost immediately because that requirements file has accumulated quite a collection of conflicting dependencies). There's [a manual trigger option](https://github.com/pyinstaller/pyinstaller-hooks-contrib/#run-the-test-on-cicd) which we used to use. [I've triggered it](https://github.com/bwoodsend/pyinstaller-hooks-contrib/actions/runs/2652778556) for you. All passing so far.

rokm:
The test pipeline is probably installing only the part of `requirements-test-libraries.txt` that has been changed, so PyQt5 is not included in that. 

As the test implicitly depends on `PyQt5` being installed (which is not automatically installed by installing `pyqtgraph`), it should probably be also guarded by an import check for `PyQt5`. In fact, the test should also be excluding all other PySide/PyQt variants from the build, because if those happen to be installed in the test environment, they will all be collected (due to references in the collected templates), which will result in undefined behavior. But considering our CI currently installs only `PyQt5`, we can probably ignore that right now...

nicoddemus:
Hi folks, thanks for the quick response!

> [I've triggered it](https://github.com/bwoodsend/pyinstaller-hooks-contrib/actions/runs/2652778556) for you. All passing so far.

Thanks, it seems everything is passing.

> As the test implicitly depends on PyQt5 being installed (which is not automatically installed by installing pyqtgraph), it should probably be also guarded by an import check for PyQt5

Actually it will work with any variant supported by `pyqtgraph`, although I don't know what would happen if we have multiple variants installed at the same time...

Let me know if you would like any other changes, and I will be glad to apply them. 👍

bwoodsend:
Can you squirt something like the following line in [here](https://github.com/pyinstaller/pyinstaller-hooks-contrib/blob/a3ddd1d597834733e5555e3537cdd5080958ee85/.github/workflows/pr-test.yml#L40):

```bash
if grep -q pyqtgraph requirements-test-libraries.txt ; then echo PyQt5 >> requirements-test-libraries.txt ;fi
```

That way, PyQt5 will be installed on this PR and any PyUp-date PR that bumps pyqtgraph. Then, like Rok says, can we have an `@importorskip("PyQt5")` on that new test.

nicoddemus:
@bwoodsend done!

nicoddemus:
Nice, it worked. 👍 

```
src/_pyinstaller_hooks_contrib/tests/test_libraries.py::test_pyqtgraph[onedir] PASSED [ 99%]
src/_pyinstaller_hooks_contrib/tests/test_libraries.py::test_pyqtgraph[onefile] PASSED [100%]
```

rokm (approved):

bwoodsend:
Hrmm, CI/CD still stalls on Ubuntu. I thought we had xvfb and the Qt dependencies already set up for this workflow. Unless anyone has any idea why, I'm just going to merge this anyway since it passes locally?

rokm:
We're [missing](https://github.com/rokm/pyinstaller-hooks-contrib/runs/7308898671?check_suite_focus=true#step:11:21) `libxcb-image0`. Which in turn causes a freeze in `collect_submodules` when one of `pyqtgraph` sub-packages tries to load the xcb plugin (presumably it tries to instantiate `QApplication`).

Installing the missing package [fixes the problem](https://github.com/rokm/pyinstaller-hooks-contrib/runs/7308941068?check_suite_focus=true).

bwoodsend:
Ahh, cool. We'll cherry pick that missing package addition separately.

bwoodsend:
Thanks for the help @nicoddemus!

nicoddemus:
Sure thing, thanks folks for maintaining this package!

PR #471, opened 2022-07-25
hooks: fix hook-pypylon for PyInstaller 5.1

nyyManni:
The hook was broken in Pyinstaller commit:
https://github.com/pyinstaller/pyinstaller/commit/20f922cdac0207e943395bcae4a23d597558c5f7

Before that, the hook worked "by accident".

### DIFF
--- a/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-pypylon.py
+++ b/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-pypylon.py
@@ -22,7 +22,6 @@
 
 import os
 
-from PyInstaller.utils.hooks import collect_all
 from PyInstaller.utils.hooks import collect_data_files
 from PyInstaller.utils.hooks import collect_dynamic_libs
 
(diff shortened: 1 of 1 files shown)


src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-pypylon.py
```diff
@@ -35,7 +35,7 @@
 # Exclude the C++-extensions from automatic search, add them manually as data files
 # their dependencies were already handled with collect_dynamic_libs
 excludedimports = ['pypylon._pylon', 'pypylon._genicam']
-for filename, module in collect_all('pypylon._pylon')[0]:
+for filename, module in collect_all('pypylon')[0]:
```
rokm, line 38: As we are interested only in datas list (the first element of the returned tuple), this `collect_all` call should be replaced with `collect_data_files('pypylon', include_py_files=True)˙.

nyyManni: You are absolutely right. Made the change.

bwoodsend:
Thank you!

PR #475, opened 2022-08-14
hooks: update pywintypes and pythoncom hooks

rokm:
Have `pywintypes` and `pythoncom` hook collect their corresponding DLLs into `pywin32_system32` sub-directory if the DLLs originally came from `site-packages/pywin32_system32` directory (which is the case with PyPI wheel, but not with Anaconda package). This prevents duplication of `pywintypesXY.dll` when the latter is also collected as a part of link-time binary dependency analysis, which starting with PyInstaller 5.4 attempts to preserve the parent directory structure (https://github.com/pyinstaller/pyinstaller/pull/7028).

### DIFF
--- a/news/474.update.rst
+++ b/news/474.update.rst
@@ -1,0 +1,2 @@
+Update ``pywintypes`` and ``pythoncom`` hooks for compatibility with upcoming
+changes in PyInstaller's attempt at preserving DLL parent directory structure.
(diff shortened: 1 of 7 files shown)


rokm:
As a side note, the changes here are essentially no-op (e.g., in CI tests) until we bump the version of PyInstaller in `develop` to 5.4.

bwoodsend (approved):

PR #490, opened 2022-09-26
Update hook-pyqtgraph.py

carl-andersson:
Pyqtgraph is missing a data include for CSV files which are needed for colormaps


rokm:
Hmm, there's also `.hex` files (e.g., `pyqtgraph/colors/maps/PAL-relaxed.hex`) and `.txt` files that contain licensing information (and should almost certainly be collected along the `.csv` files).

Perhaps instead of trying to explicitly include all possible data files, we should be just excluding the ones that we do not want - presumably the `.cfg` files from `pyqtgraph/examples`:
```python
datas = collect_data_files('pyqtgraph', excludes=['**/examples/*'])
```

carl-andersson:
The change I submitted is sufficient and working and I do not have enough knowledge of pyqtgraph to have any input on the change you suggest. I just figured that someone else might run into the same problem

rokm:
Superseded by #501.

PR #501, opened 2022-10-19
Update hooks for pyqtgraph

rokm:
Update `pyqtgraph` standard hook to collect all data from the package, including the colormaps and their associated license files, as per https://github.com/pyinstaller/pyinstaller-hooks-contrib/pull/490#issuecomment-1258173167.

Create a new runtime hook that attempts to handle subprocess spawned by `pyqtgraph.multiprocess`. The caveat is that the user still needs to call stdlib `multiprocess.freeze_support` in their entry-point script (which is something I hope to get around to addressing in PyInstaller's `multiprocess` hook). And for onefile builds, they also need to set `_MEIPASS2` environment variable to prevent the executable from unpacking itself again in the subprocess (using different `sys._MEIPASS`, and thus breaking the path check in the runtime hook). This is because `pyqtgraph.multiprocess` uses `subprocess.Popen` to spawn the process; perhaps we could extend our `subprocess`  runtime hook to detect when `sys.executable` is spawned and automatically restore `_MEIPASS2` in the subprocess' environment?

Replaces and closes #490.

### DIFF
--- a/src/_pyinstaller_hooks_contrib/tests/test_libraries.py
+++ b/src/_pyinstaller_hooks_contrib/tests/test_libraries.py
@@ -1289,6 +1289,16 @@ def test_pyqtgraph(pyi_builder):
     )
 
 
+@importorskip('pyqtgraph')
+def test_pyqtgraph_colormap(pyi_builder):
+    pyi_builder.test_source(
+        """
+        import pyqtgraph.colormap
+        assert pyqtgraph.colormap.listMaps()
+        """
+    )
+
+
 @importorskip('hydra')
 def test_hydra(pyi_builder, tmpdir):
     config_file = str((Path(__file__) / '../data/test_hydra/config.yaml').resolve(strict=True).as_posix())
--- a/src/_pyinstaller_hooks_contrib/tests/test_libraries.py
+++ b/src/_pyinstaller_hooks_contrib/tests/test_libraries.py
@@ -1285,7 +1285,8 @@ def test_pyqtgraph(pyi_builder):
         import pyqtgraph.graphicsItems.PlotItem
         import pyqtgraph.graphicsItems.ViewBox.ViewBoxMenu
         import pyqtgraph.imageview.ImageView
-        """
+        """,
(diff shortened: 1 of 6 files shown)


rokm:
Here's oneshot test for the added tests: https://github.com/rokm/pyinstaller-hooks-contrib/actions/runs/3284554698

On linux, the oneshot workflow lacks the apt-installed packages required by `PyQt5` and lacks `Xvfb` setup, hence the failures.

src/_pyinstaller_hooks_contrib/tests/test_libraries.py
```diff
@@ -1299,6 +1300,47 @@ def test_pyqtgraph_colormap(pyi_builder):
...
+        import pyqtgraph
+
+        # Multiprocessing is used internally by pyqtgraph.multiprocess
+        import multiprocessing
+        multiprocessing.freeze_support()
+
+        # pyqtgraph.multiprocess also uses a subprocess.Popen() to spawn its
+        # sub-process, so we need to restore _MEIPASS2 to prevent the executable
+        # to unpacking itself again in the subprocess.
+        os.environ['_MEIPASS2'] = sys._MEIPASS
```
bwoodsend, line 1322: Any reason why this can't go in a runtime hook?

rokm: Yeah, it would potentially cause issues when trying to run a different PyInstaller-frozen executable from the program, which is the main reason why the environment variable is cleared in the first place.

I think the only way to handle this automatically in the safe way is if we manage to patch `subprocess.Popen` to detect that it is trying to run `sys.executable`.

rokm: > Yeah, it would potentially cause issues when trying to run a different PyInstaller-frozen executable from the program, which is the main reason why the environment variable is cleared in the first place.

Sure, the same problem arises if the user sets the environment variable in their code, but *they* should know whether their program will also try to spawn a subprocess with a different pyinstaller-frozen executable or not.

PR #505, opened 2022-11-02
hooks: pyproj: collect metadata for pyproj >= 3.4.0

rokm:
As of version 3.4.0, pyproj's `__init__.py` queries its own version via `importlib.metadata.version(__package__)`, which requires the package metadata to be collected.

### DIFF
--- a/news/505.update.rst
+++ b/news/505.update.rst
@@ -1,0 +1,1 @@
+Update ``pyproj`` hook for compatibility with ``pyproj`` v3.4.0.
--- a/requirements-test-libraries.txt
+++ b/requirements-test-libraries.txt
@@ -25,7 +25,7 @@ fabric==2.7.1
 fiona==1.8.22; sys_platform != 'win32'
 folium==0.13.0
 ffpyplayer==4.3.5; python_version < '3.10' # doesn't have py310 wheels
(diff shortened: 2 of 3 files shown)


bwoodsend:
They did revert it (https://github.com/pyproj4/pyproj/pull/1156). That's why I was putting off doing anything about it. I was hoping they'd release 3.4.1 then we could pretend 3.4.0 never happened.

rokm:
Ah, fair enough. Let's ignore 3.4, then.

rokm:
If pyproj 3.4.1 is happening on December 1st, this means we have a whole-month-long window in which we'll have to be dealing with bug reports about this.

Therefore, I'm reopening this PR, but I've changed the version check from `>= 3.0.0` to `== 3.4.0` to limit metadata collection only to the problematic release.

We will need to push out new hooks-contrib release, though...

bwoodsend:
Go ahead and hit the [release button](https://github.com/pyinstaller/pyinstaller-hooks-contrib/actions/workflows/release.yml) after merging.

rokm:
I'll try to find out what the deal with double headings in ChangeLog is, and then make a release.

bwoodsend:
It'll be the towncrier template again. This is at least the 4th time they've changed their minds on whether the headings should be specified in the template or inserted automatically.

PR #508, opened 2022-11-03
Add hook and test for exchnagelib

arossert:
In some cases `exchengelib` need to collect `tzdata`, I noticed it on `alpine` distribution that it raises this on import
```
exchangelib.errors.UnknownTimeZone: No time zone found with key UTC
```

### DIFF
--- a/news/508.new.rst
+++ b/news/508.new.rst
@@ -1,0 +1,1 @@
+Add hook for ``exchangelib``.
--- a/requirements-test-libraries.txt
+++ b/requirements-test-libraries.txt
@@ -111,6 +111,7 @@ pyshark==0.5.3
 opencv-python==4.6.0.66
 hydra-core==1.2.0
 spiceypy==5.1.2
+exchangelib==4.8.0
 
 # ------------------- Platform (OS) specifics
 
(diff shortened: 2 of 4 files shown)


rokm:
Wouldn't a hidden import suffice?

arossert:
> Wouldn't a hidden import suffice? EDIT: I mean a single hidden import of `tzdata`, as opposed to all its submodules.

I tested it and it seems to work with just the hidden import, I will fix the PR.

bwoodsend:
Can you add the [news entry](https://github.com/pyinstaller/pyinstaller-hooks-contrib#add-a-news-entry) and add exchangelib to [the test requirements](https://github.com/pyinstaller/pyinstaller-hooks-contrib#pin-the-test-requirement).

arossert:
> Can you add the [news entry](https://github.com/pyinstaller/pyinstaller-hooks-contrib#add-a-news-entry) and add exchangelib to [the test requirements](https://github.com/pyinstaller/pyinstaller-hooks-contrib#pin-the-test-requirement).

Done, one question, why fix the versions in the requirements file? Don't we want to check if the hook still works on the latest version?

rokm (approved):

rokm:
> Done, one question, why fix the versions in the requirements file? Don't we want to check if the hook still works on the latest version?

We do, but we also want the tests to be repeatable and predictable. Having the package versions pinned (and periodically updated) decouples test failures caused by changes in hooks and test failures caused by new package releases.

arossert:
Thanks for merging this, what is the general release process? when should we expect a new release?

bwoodsend:
It's normally someone reminds me that I haven't released this for a while then I hit the release button. Said button has now been pressed.

PR #520, opened 2022-11-22
bugfix: Unable to find "/usr/sbin/neato" when adding binary and data …

jan-fuchs-fuky:
…files.

Problem on Debian GNU/Linux 11 (bullseye) and Ubuntu 22.04.1 LTS:

    In : os.path.realpath(shutil.which("dot"))
    Out: '/usr/sbin/libgvc6-config-update'

    In : os.path.dirname(os.path.realpath(shutil.which("dot")))
    Out: '/usr/sbin'

Solution:

    In : os.path.dirname(shutil.which("dot"))
    Out: '/usr/bin'

### DIFF
--- a/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-pygraphviz.py
+++ b/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-pygraphviz.py
@@ -49,7 +49,7 @@ if is_win:
 else:
     # The dot binary in PATH is typically a symlink, handle that.
     # graphviz_bindir is e.g. /usr/local/Cellar/graphviz/2.46.0/bin
-    graphviz_bindir = os.path.dirname(os.path.realpath(shutil.which("dot")))
+    graphviz_bindir = os.path.dirname(shutil.which("dot"))
     for binary in progs:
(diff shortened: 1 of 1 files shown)


bwoodsend:
Indeed, `dot` is a symlink so we want to collect the file it points to but still call it `dot`.
```
root@924c75cb9e42:/# stat /usr/bin/dot                   
  File: /usr/bin/dot -> ../sbin/libgvc6-config-update
```
[The comment just above](https://github.com/pyinstaller/pyinstaller-hooks-contrib/pull/520/files#diff-19744a4a9355fb0219cc8de47560ce33f7f86ff29c32a63c9d357493ef15f56eL50-L51) however points out that on macOS, you need to resolve the symlink to get to the real graphviz installation hiding in homebrew's cellar.

rokm:
For the executables, wouldn't it be safer if we just resolved each and every one using `shutil.which` instead of trying to resolve their directory and collecting them from it?

lcbw:
I'm looking to leverage this capability. Has there been any movement on this PR?

bwoodsend:
> Has there been any movement on this PR?

Does it look like there has?

PR #530, opened 2022-12-30
hooks: add hook for easyocr

Jmennius:
I hope what I've done is not too controversial :)

### DIFF
--- a/hooks-config.rst
+++ b/hooks-config.rst
@@ -1,0 +1,53 @@
+===========================
+Supported hooks and options
+===========================
+
+This section lists hooks that are configurable in `pyinstaller-hooks-contrib`.
+
+See also `Hook Configuration Options <https://pyinstaller.org/en/stable/hooks-config.html>`_
+for more information regarding hook options - how to add new hook options,
+hook options from primary pyinstaller repo, etc.
+
+
+easyocr hook
+------------
+
+easyocr hook allows configuration of included language-related data files
+(there are a number of text files with characters and dictionaries per language).
+Dictionaries are quite large (about 290MB at the time of writing),
+so it's likely that end-users would want to only include what they use.
+
+If not set to languages used, an error in `setLanguageList()` will occur at runtime::
+
+   No such file or directory: '<PATH>/easyocr/character/<LANG_CODE>_char.txt'
+
+
+**Hook identifier:** ``easyocr``
+
+**Options**
+
+ * ``lang_codes`` [*list of strings*]: list of ISO 639 language code (e.g., ``en``) for
+   which data files should be collected.
+   By default, all languages will be included (``*`` in place of the language code).
+
+**Example**
+
+Collect only data files required for English, French and Spanish::
+
+    a = Analysis(
+        ['my-ocr-app.py'],
+        ...,
+        hooksconfig={
+            'easyocr': {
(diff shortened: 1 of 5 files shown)


bwoodsend:
Ahh phew, you had me expecting `collect_all()` used in a for loop plus some hard-coded brew paths and perhaps some cuda dlls.

We do already have a mechanism for hooks with parameters. See [here](https://pyinstaller.org/en/stable/hooks-config.html) for the about and [here](https://pyinstaller.org/en/stable/hooks-config.html#adding-an-option-to-the-hook) for adding options. Since you're the first to use it in this repo, you'll need to create a new *Supported hooks options* markdown file to put the documentation in. We'll later need to add a link to it in https://pyinstaller.org/en/stable/hooks-config.html#supported-hooks-and-options.

Jmennius:
> Ahh phew, you had me expecting `collect_all()` used in a for loop plus some hard-coded brew paths and perhaps some cuda dlls.
> 
> We do already have a mechanism for hooks with parameters. See [here](https://pyinstaller.org/en/stable/hooks-config.html) for the about and [here](https://pyinstaller.org/en/stable/hooks-config.html#adding-an-option-to-the-hook) for adding options. Since you're the first to use it in this repo, you'll need to create a new _Supported hooks options_ markdown file to put the documentation in. We'll later need to add a link to it in https://pyinstaller.org/en/stable/hooks-config.html#supported-hooks-and-options.

Done.
Feel free to rephrase/move or otherwise fix the doc. I've kept the hook-options doc ReST and it's in the root of the repo... 🤷🏼‍♂️

Jmennius:
I think there is another part that is quite crucial for this package - it downloads NN models (at runtime).
So I probably have to add it. Not sure when - maybe another PR, but definitely not this week :)

bwoodsend:
> I think there is another part that is quite crucial for this package - it downloads NN models (at runtime).

That depends on where it puts them and what it needs to do it. If it writes to somewhere sensible like `~/.cache` and doesn't require any dependencies that PyInstaller isn't already able to detect then we should be able to leave it alone. If it downloads into its own `site-packages` location then it'll error out under PyInstaller when someone installs to someone owned by root.

hooks-config.rst
```diff
@@ -0,0 +1,54 @@
+===========================
+Supported hooks and options
+===========================
+
+This section lists hooks that implement support for configuration
+options in `pyinstaller-hooks-contrib` repository.
```
bwoodsend, line 6:
```suggestion
This section lists hooks that are configurable in `pyinstaller-hooks-contrib`.
```

hooks-config.rst
```diff
@@ -0,0 +1,54 @@
...
+
+Please refer to `Hook Configuration Options <https://pyinstaller.org/en/stable/hooks-config.html>`_
```
bwoodsend, line 8:
```suggestion
See also `Hook Configuration Options <https://pyinstaller.org/en/stable/hooks-config.html>`_
```

Jmennius:
@bwoodsend I've squashed your suggestions!

bwoodsend (approved):
Looks good to me. I'd be tempted to ask for a test too but given that easyocr depends on pytorch+cuda, I think I'd rather not go there.

rokm (approved):

bwoodsend:
Nice one!

Issue #375, opened 2022-01-31
can't find torchtext dylib files


inferense:
Pyinstaller builds successfully, however when I run the exe I get an error related to missing torchtext dylib files.

I get the following warnings during the freezing:

```
98346 WARNING: Cannot find path ./libc++.1.dylib (needed by .../env/lib/python3.9/site-packages/torchtext/_torchtext.so)
98346 WARNING: Cannot find path ./libtorch_python.dylib (needed by .../env/lib/python3.9/site-packages/torchtext/_torchtext.so)
98346 WARNING: Cannot find path ./libtorch.dylib (needed by .../env/lib/python3.9/site-packages/torchtext/_torchtext.so)
98346 WARNING: Cannot find path ./libtorch_cpu.dylib (needed by .../env/lib/python3.9/site-packages/torchtext/_torchtext.so)
98347 WARNING: Cannot find path ./libc10.dylib (needed by .../env/lib/python3.9/site-packages/torchtext/_torchtext.so)

```

I can find the last 4 dylib files (except `libc++.1.dylib`) in my env/torch module and can copy them manually to the dist folder but that doesn't seem like a good workaround and I still can't find `libc++.1.dylib`. Updating path / binaries in .spec file doesn't help.

Any ideas on how to resolve this?

* Output of `pyinstaller --version`: ```4.8```
* Version of Python: ```3.9```
* torchtext: `0.11.2`
* torch: `1.10.2` 
* Platform: ```MacOS 12.1``` -->
* How you installed Python: python.org

geoffr98:
Romk,

Can you please explain further your comment : " The only way to get the program working is to replace the libraries in top-level directory with symbolic links"

What top-level directory?

I am asking because I think I'm seeing the same problem. When I run pyinstaller I see the following warnings:
`37358 WARNING: Cannot find path ./libtorch_cpu.dylib (needed by /Users/rpm/.pyenv/versions/3.8.8/lib/python3.8/site-packages/torchaudio/_torchaudio.so)
37358 WARNING: Cannot find path ./libtorchaudio.so (needed by /Users/rpm/.pyenv/versions/3.8.8/lib/python3.8/site-packages/torchaudio/_torchaudio.so)
37358 WARNING: Cannot find path ./libc10.dylib (needed by /Users/rpm/.pyenv/versions/3.8.8/lib/python3.8/site-packages/torchaudio/_torchaudio.so)
[...]

PR #531, opened 2022-12-30
hooks: torch: Only include shared libraries

Jmennius:
Existing hook is way too simplistic and literally takes every single file in the package (headers, cpp, pyi, __pycache__ and so on)...
Fix this by only bundling shared libraries from `lib` dir.

This depends on pyinstaller/pyinstaller#7357.
Please advise how we can specify dependency on main repository (if we should).

P.S. Need to test this one more time (it was some time ago I've implemented this).

### DIFF
--- a/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-torch.py
+++ b/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-torch.py
@@ -10,6 +10,7 @@
 # SPDX-License-Identifier: GPL-2.0-or-later
 # ------------------------------------------------------------------
 
-from PyInstaller.utils.hooks import get_package_paths
+from PyInstaller.utils.hooks import collect_dynamic_libs
 
-datas = [(get_package_paths('torch')[1],"torch"),]
+# include versioned .so files as well
+binaries = collect_dynamic_libs('torch', py_dylib_patterns=['*.dll', '*.dylib', 'lib*.so', 'lib*.so.*'])


bwoodsend:
> Existing hook is way too simplistic and literally takes every single file in the package (headers, cpp, pyi, pycache and so on)...

For the headers and cpp files, I'd argue that torch's packaging is too simplistic - they shouldn't have found their way onto PyPI for PyInstaller to then need to clear up.

>  Please advise how we can specify dependency on main repository (if we should).

I think for this, just do an

```python
if is_module_satisfies('pyinstaller >= 5.8'):
    new behaviour
else:
    old behaviour
```

> P.S. Need to test this one more time (it was some time ago I've implemented this).

If you add it to `requirements-test-libraries.txt` then CI will take care of that.

rokm:
Long-term, this is definitely a way to go (this horrible brute-force hook needs to be rewritten), but at this point, I suspect this is going to cause several regressions:

1) while dynamic libs definitely should be collected as binaries, doing so in this cause will cause duplication on macOS, which will in turn cause issues similar to #375 even for people that use just `torch` instead of `torchaudio` and/or `torchvideo`

2) I strongly suspect we should brute-force collect the `torch` sub-modules via `collect_submodules` as well.

3) I also suspect we need to collect the source .py files (for JIT), but instead of the current brute-force approach, we can leverage `collect_submodules` + package collection mode (#6945), i.e., 'pyz+py'. This should ensure that all collected modules are also analyzed for their imports.

However, to avoid issues with 1), we first need to add support for symlinks to PyInstaller (so we can avoid duplication by keeping collected dylibs in their original location and symlinking them to _MEIPASS to satisfy PyInstaller's current assumption that they are collected to _MEIPASS).

So my plan was to rewrite this hook once we get symlinks done, using the afore-mentioned `if is_module_satisfies('pyinstaller >= 5.X')` switch to keep the old brute-force behavior for older versions of PyInstaller.

Jmennius:
> Long-term, this is definitely a way to go (this horrible brute-force hook needs to be rewritten), but at this point, I suspect this is going to cause several regressions:
> 
>     1. while dynamic libs definitely should be collected as binaries, doing so in this case will cause duplication on macOS, which will in turn cause issues similar to [can't find torchtext dylib files #375](https://github.com/pyinstaller/pyinstaller-hooks-contrib/issues/375) even for people that use just `torch` instead of `torchaudio` and/or `torchvideo`
> 
>     2. I strongly suspect we should brute-force collect the `torch` sub-modules via `collect_submodules` as well.
> 
>     3. I also suspect we need to collect the source .py files (for JIT), but instead of the current brute-force approach, we can leverage `collect_submodules` + package collection mode ([Implement module collection mode setting for hooks pyinstaller#6945](https://github.com/pyinstaller/pyinstaller/pull/6945)), i.e., 'pyz+py'. This should ensure that all collected modules are also analyzed for their imports.
> 
> 
> However, to avoid issues with 1), we first need to add support for symlinks to PyInstaller (so we can avoid duplication by keeping collected dylibs in their original location and symlinking them to _MEIPASS to satisfy PyInstaller's current assumption that they are collected to _MEIPASS).
> 
> So my plan was to rewrite this hook once we get symlinks done, using the afore-mentioned `if is_module_satisfies('pyinstaller >= 5.X')` switch to keep the old brute-force behavior for older versions of PyInstaller.

Ah, you mentioning JIT reminds me of something... maybe that is why I did not push it. (that's why I wanted to test it first at runtime 😓)
I will close this one for now, until either one of use can return and finish it (I can't at the moment + there are prerequisites).
Thanks!

PR #542, opened 2023-02-06
Add hook for customtkinter

JakobDev:
This adds a hook for [customtkinter](https://pypi.org/project/customtkinter/)

### DIFF
--- a/news/542.new.rst
+++ b/news/542.new.rst
@@ -1,0 +1,1 @@
+Add hook for ``customtkinter``
--- a/requirements-test-libraries.txt
+++ b/requirements-test-libraries.txt
@@ -124,6 +124,7 @@ NBT==1.5.1
 minecraft-launcher-lib==5.3; python_version >= '3.8'
 scikit-learn==1.2.2; python_version >= '3.8'
 scikit-image==0.20.0; python_version >= '3.8'
+customtkinter==5.1.3
 
 # ------------------- Platform (OS) specifics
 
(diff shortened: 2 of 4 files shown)


rokm:
>```
>FileNotFoundError: [Errno 2] No such file or directory: '/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/_MEIrh5Ytq/customtkinter/windows/widgets/theme/../../../assets/themes/blue.json'
>```
Looks like `customtkinter` will need to be patched for compatibility with PyInstaller on POSIX systems.

In the frozen application, python modules are by default collected to executable-embedded PYZ archive, so package sub-directories that contain only python modules do not end up collected. This is why in this example, `customtkinter/windows/widgets/theme` does not exist. And on POSIX systems, you cannot have paths relative to non-existing directories, so `customtkinter/windows/widgets/theme/../../../assets/themes/blue.json` results in file-not-found, even though `customtkinter/assets/themes/blue.json` does exist.

Therefore, this [attempt at reading the file](https://github.com/TomSchimansky/CustomTkinter/blob/59df37e920fc01c5cce5bc9aacee0989048e91c1/customtkinter/windows/widgets/theme/theme_manager.py#L18) should normalize the path using `os.normpath` to turn `customtkinter/windows/widgets/theme/../../../assets/themes/blue.json` into `customtkinter/assets/themes/blue.json`.

bwoodsend (approved):

rokm:
Looks like the issue with relative paths has been resolved, so this should now pass tests on all OSes, and can be merged. Thanks @JakobDev!

PR #545, opened 2023-02-13
Add work-around for ffpyplayer trying to use site.USER_BASE

rokm:
Add a run-time hook for `ffpyplayer` that sets site.USER_BASE to an empty string, in order to prevent run-time errors caused by `ffpyplayer` blindly passing `site.USER_BASE` to `os.path.join`.

Due to `site` being disabled in PyInstaller-frozen application, `site.USER_BASE` is `None`. In PyInstaller versions prior to v5.5, this was mitigated by PyInstaller's own fake `site` module, which also set `site.USER_BASE` to an empty string.

### DIFF
--- a/news/545.update.rst
+++ b/news/545.update.rst
@@ -1,0 +1,3 @@
+Add work-around for ``ffpyplayer`` 4.3.5 and 4.4.0 trying to use
+``site.USER_BASE``, which is ``None`` in  PyInstaller 5.5 and later
+due to removal of PyInstaller's fake ``site`` module.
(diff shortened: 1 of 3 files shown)


rokm:
Oneshot tests: https://github.com/rokm/pyinstaller-hooks-contrib/actions/runs/4167890121

bwoodsend:
I've also sent off https://github.com/matham/ffpyplayer/pull/141 but given that almost every other case where I've taken the initiative to fix the problem upstream has led to a PR sitting rotting for all eternity, I won't defer patching it over our side.

bwoodsend (approved):

PR #546, opened 2023-02-13
Add work-around for tensorflow < 2.3.0 trying to use site.USER_SITE

### LINKED ISSUE #529: Upstream removal of fake `site` module breaks TensorFlow support
**Describe the bug**
PyInstaller's fake `site` module was [recently removed](https://github.com/pyinstaller/pyinstaller/commit/7bc99716a6b8db9c31b3a8e84fe708d29db53444#diff-cbf12f30c55dde378d38d8fba5d9bb09d5c163ea09c5b40b4341b224a14c33a0), allowing the value of `site.USER_SITE` to once again be `None`. This bug affects PyInstaller from 5.5 through 5.7

This causes a problem with some [introspection](https://github.com/tensorflow/tensorflow/blob/4cbd1d017e45c2d1cb5aa3011a569e6f3b950aa2/tensorflow/api_template.__init__.py#L125-L147) in TensorFlow that runs at import time which assumes this value will always be a string.

Users affected by this bug can either pin `PyInstaller<5.5` or work around the issue by executing `import site; site.USER_SITE = ''` before any TensorFlow imports occur (either in your application code or in a custom runtime hook), and the fix is probably to do something similar in a runtime hook for the library.

**To Reproduce**

A minimal example file:
```
import tensorflow.lite

print("Test program: tflite import succeeded")
```

PyInstaller command:
```
python -m PyInstaller --collect-all tensorflow testapp.py
```

Error (when running the resulting distribution):
```
PS C:\> .\dist\testapp\testapp.exe
  File "testapp.py", line 1, in <module>
    import tensorflow.lite
  File "<frozen importlib._bootstrap>", line 983, in _find_and_load
  File "<frozen importlib._bootstrap>", line 967, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 677, in _load_unlocked
  File "PyInstaller\loader\pyimod02_importers.py", line 499, in exec_module 
  File "tensorflow\__init__.py", line 102, in <module>
  File "<frozen importlib._bootstrap>", line 983, in _find_and_load
  File "<frozen importlib._bootstrap>", line 967, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 677, in _load_unlocked
  File "PyInstaller\loader\pyimod02_importers.py", line 499, in exec_module 
  File "tensorflow_core\__init__.py", line 771, in <module>
  File "tensorflow_core\__init__.py", line 769, in _running_from_pip_package
  File "tensorflow_core\__init__.py", line 769, in <genexpr>
TypeError: startswith first arg must be str or a tuple of str, not NoneType 
[15524] Failed to execute script 'testapp' due to unhandled exception!
```

Note that the line numbers reported here are on the far side of template interpolation during TensorFlow's build process, `tensorflow_core\__init__.py:769` in this error report corresponds to [Line 130 of the templated file](https://github.com/tensorflow/tensorflow/blob/4cbd1d017e45c2d1cb5aa3011a569e6f3b950aa2/tensorflow/api_template.__init__.py#L130)

**Expected behavior**
The test program should not error on the import and should print the success message, as it does in PyInstaller versions preceding 5.5

**Screenshots**
N/A

 - OS: Windows 10
 - Python Version: 3.7.9
 - Version of `pyinstaller-hooks-contrib`: 2022.14
 - Version of PyInstaller: 5.7

**Additional context**
N/A

rokm:
Add a run-time hook for `tensorflow` that sets site.USER_SITE to an empty string, in order to prevent run-time errors caused by `tensorflow` < 2.3.0 blindly passing `site.USER_SITE` to `str.startswith`.

Due to `site` being disabled in PyInstaller-frozen application, `site.USER_SITE` is `None`. In PyInstaller versions prior to v5.5, this was mitigated by PyInstaller's own fake `site` module, which also set `site.USER_SITE` to an empty string.

Fixes #529.

### DIFF
--- a/news/546.update.rst
+++ b/news/546.update.rst
@@ -1,0 +1,3 @@
+Add work-around for ``tensorflow`` < 2.3.0 trying to use
+``site.USER_SITE``, which is ``None`` in  PyInstaller 5.5 and later
+due to removal of PyInstaller's fake ``site`` module.
--- a/src/_pyinstaller_hooks_contrib/hooks/rthooks.dat
+++ b/src/_pyinstaller_hooks_contrib/hooks/rthooks.dat
@@ -9,4 +9,5 @@
     'pythoncom': ['pyi_rth_pythoncom.py'],
     'pyqtgraph': ['pyi_rth_pyqtgraph_multiprocess.py'],
     'pywintypes': ['pyi_rth_pywintypes.py'],
+    'tensorflow': ['pyi_rth_tensorflow.py'],
 }
--- a/src/_pyinstaller_hooks_contrib/hooks/rthooks/pyi_rth_tensorflow.py
+++ b/src/_pyinstaller_hooks_contrib/hooks/rthooks/pyi_rth_tensorflow.py
@@ -1,0 +1,19 @@
+#-----------------------------------------------------------------------------
+# Copyright (c) 2023, PyInstaller Development Team.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+#
+# The full license is in the file COPYING.txt, distributed with this software.
+#
+# SPDX-License-Identifier: Apache-2.0
+#-----------------------------------------------------------------------------
+
+# `tensorflow` versions prior to 2.3.0 attempt to use `site.USER_SITE` in path/string manipulation functions.
+# As frozen application runs with disabled `site`, the value of this variable is `None`, and causes path/string
+# manipulation functions to raise an error. As a work-around, we set `site.USER_SITE` to an empty string, which is
+# also what the fake `site` module available in PyInstaller prior to v5.5 did.
+import site
+
+if site.USER_SITE is None:
+    site.USER_SITE = ''


bwoodsend (approved):
Hmm, I'd prefer to close #529 as _won't fix_ but if you think it's worth it...

rokm:
The problem with `tensorflow` is that some methods/models are implemented in old versions of `tensorflow` and cannot be easily upgraded to newer versions of `tensorflow`. 

I have misfortune of having a project that's stuck on `tensorflow` 2.2.0 and uses PyInstaller to produce Windows binaries, so this is mostly me trying to make my life easier next time I have to rebuild it... :)

SnoopJ:
> I have misfortune of having a project that's stuck on `tensorflow` 2.2.0 and uses PyInstaller to produce Windows binaries, so this is mostly me trying to make my life easier next time I have to rebuild it... :)

This is basically the context that caused me to file the issue in the first place, although we're pinned to an _even older_ TensorFlow. I have a workaround in place, but look forward to removing it in the future whenever we resolve some unrelated our-side conflicts with newer versions of PyInstaller.

I agree that it's an edge case, but thanks for putting in a patch for it!

PR #551, opened 2023-02-27
hooks: prevent pyqtgraph hook from recursing into pyqtgraph.examples

rokm:
Prevent `pyqtgraph` hook from recursing into `pyqtgraph.examples` while scanning for submodules. While the obtained list of submodules is subsequently filtered to include only templates, import of `pyqtgraph.examples` during scan triggers instantiation of `QtWidgets.QApplication`. This, in turn, requires a valid X/Wayland session on linux and and lack thereof results in an error that crashes the python subprocess in which we scan for submodules.

Fixes the second (linux-specific) part of the issue reported in pyinstaller/pyinstaller#7452.

### DIFF
--- a/news/551.update.rst
+++ b/news/551.update.rst
@@ -1,0 +1,2 @@
+Prevent ``pyqtgraph`` hook from recursing into ``pyqgraph.examples``
+while scanning for submodules.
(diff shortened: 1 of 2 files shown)


bwoodsend (approved):

jdegenstein:
I tested this change and it resolved my issue! Thank you so much for the fast help and fix!

Issue #438, opened 2022-05-21
ModuleNotFoundError: No module named 'yt_dlp.compat._legacy'


aliencaocao:
Unable to launch compiled exe using [yt_dlp ](https://github.com/yt-dlp/yt-dlp) library. If I run the python .py file, works fine.
Was able to do it in PyInstaller 5.0.

```
Traceback (most recent call last):
  File "main.py", line 10, in <module>
    import yt_dlp
  File "PyInstaller\loader\pyimod03_importers.py", line 495, in exec_module
  File "yt_dlp\__init__.py", line 11, in <module>
  File "PyInstaller\loader\pyimod03_importers.py", line 495, in exec_module
  File "yt_dlp\compat\__init__.py", line 14, in <module>
  File "yt_dlp\compat\compat_utils.py", line 36, in passthrough_module
  File "importlib\__init__.py", line 127, in import_module
ModuleNotFoundError: No module named 'yt_dlp.compat._legacy'
[22836] Failed to execute script 'main' due to unhandled exception!
```

Code:
```
import yt_dlp
```

Compiled with `pyinstaller --log-level DEBUG -F --add-data "libopus-0.dll;." --distpath ./ --clean --upx-dir upx/ --noconfirm main.py`

bwoodsend:
You can submit a hook for it if you want. I'd much rather that `yt_dlp` stopped doing [their hairy dynamic imports hacks](https://github.com/yt-dlp/yt-dlp/blob/1d485a1a799bbeeb2faea0595676ca7d4c0f3716/yt_dlp/compat/compat_utils.py).

jlw4049:
Should I post an issue on their tracker?

bwoodsend:
Hmm, you can if you want. I'd be surprised if they got rid of it just for our benefit.

jlw4049:
So should we, for now just use the command line 
--hidden-import=yt_dlp.compat._legacy

pukkandan:
The "hairy dynamic imports hacks" is necessary for us to be able to deprecate (and issue warnings for) legacy functions without actually breaking the extractor API. I have now made yt-dlp handle the import failure of `compat._legacy` gracefully. So the hook is no longer needed as long as you don't actually try to access a legacy function.

But this got me wondering - is there any way to specify in yt-dlp itself that the package has hidden imports, so that hooks don't need to be maintained separately?

bwoodsend:
[...]

PR #556, opened 2023-03-14
Remove yt_dlp hook and test

bwoodsend:
See https://github.com/pyinstaller/pyinstaller-hooks-contrib/issues/438#issuecomment-1468438175.


rokm:
I'd prefer to keep our copy around as a fallback for users of older versions of `yt-dlp` that do not provide the hook, but on the other hand, our version will shadow the upstream one for anyone not using PyInstller 5.9 (resulting in breakage for those combinations, since upstream hook is needed).

So everything considered, it is probably is indeed best to remove our copy...

rokm:
We should also remove `yt-dlp` entry from `requirements-test-libraries.txt˙ (or leave the test in).

bwoodsend:
> We should also remove `yt-dlp` entry from `requirements-test-libraries.txt` (or leave the test in).

Grr, it escaped by grep due to the `-` vs `_` mismatch.

rokm (approved):

PR #564, opened 2023-03-31
Update hook-tensorflow.py

Tombana:
In the recent TF 2.12 release, there is a new file `libtensorflow_cc.so.2` which now appears twice in the final output: once correctly in the root directory (as 'binary') and once incorrectly in the `tensorflow` subdirectory (as 'data' file). This new file is where they moved the majority of the code, it's over 500 MB, so this duplication causes a lot of overhead.

To reproduce the issue, install `tensorflow==2.12` or `tensorflow-cpu==2.12` and package a file with only `import tensorflow`. With these changes the file only appears once in the final output and the package still works as expected.

I have tested this change on Linux only.

### DIFF
--- a/news/564.update.rst
+++ b/news/564.update.rst
@@ -1,0 +1,1 @@
+Update the hook for ``tensorflow`` to be compatible with TensorFlow 2.12.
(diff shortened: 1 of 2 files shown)


rokm:
> once correctly in the root directory (as 'binary') and once incorrectly in the tensorflow subdirectory (as 'data' file)

Just for the record, it is actually the other way around; because the binary dependency analysis (incorrectly) collects the copy (as binary) to the root directory, the data copy in the subdirectory needs to be suppressed. At least until we sort out preservation of binaries' subdirectory structure on all platforms.

Tombana:
Thanks for the quick response!

rokm (approved):

rokm:
Thanks @tombana!

PR #565, opened 2023-03-31
Hooks skimage fix

ryandeon:
starting skimage 0.20, I believe, many things are lazy-loaded. To get my project with skimage to work with pyinstaller, I had to upgrade lazy_loader -> 0.2 and include this hook file. hook file created by reading https://github.com/scikit-image/scikit-image/issues/6772#issuecomment-1449671566 and guessing.

Tests did not run for me. There is already a test_skimage in the suite, but `pytest -k 'test_skimage'` would either deselect all tests or (if I removed some decorators in front of test_skimage) run the function with a bunch of parameters that are not listed in the parametrize mark. I haven't used pytest before, so maybe someone else can figure it out.

### DIFF
--- a/news/565.new.rst
+++ b/news/565.new.rst
@@ -1,0 +1,1 @@
+Added hook file for scikit-image (skimage). There were existing hooks for individual submodules of skimage, but since v0.2 it needs more.
--- a/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-skimage.py
+++ b/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-skimage.py
@@ -1,0 +1,20 @@
+# ------------------------------------------------------------------
+# Copyright (c) 2020 PyInstaller Development Team.
+#
+# This file is distributed under the terms of the GNU General Public
+# License (version 2.0 or later).
+#
+# The full license is available in LICENSE.GPL.txt, distributed with
+# this software.
+#
+# SPDX-License-Identifier: GPL-2.0-or-later
+# ------------------------------------------------------------------
+
+
+# This hook was tested with scikit-image (skimage) 0.14.1:
+# https://scikit-image.org
+
+from PyInstaller.utils.hooks import collect_data_files, collect_submodules
+hiddenimports = [x for x in collect_submodules("skimage") if "tests" not in x]
+datas =collect_data_files("skimage")
+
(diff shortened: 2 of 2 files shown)


ryandeon:
I definitely understand wanting to avoid overkill. I have some reasons for it, one bad and ...well, you can judge the others.
bad one: I didn't know which submodules might be lazy_loaded or not, I haven't really peered into the internals of skimage.
reason 2) if the hook targets just the ones that are currently lazy_loaded, I imagine that with each version, as skimage updates more submodules, they will convert more and more to using this new way. So the hook will have to be updated in lockstep. I don't actually know what their roadmap is, I'm just being pessimistic about it.
reason 3) I didn't find it added much to the overall build time in pyinstaller. If there are other reasons than wasted time to avoid this kind of overkill, let me know.

ryandeon:
that being said, if you'd still prefer I pare it down to just the imports you mention, I can probably do that without a ton more effort.

rokm:
> I definitely understand wanting to avoid overkill. I have some reasons for it, one bad and ...well, you can judge the others. bad one: I didn't know which submodules might be lazy_loaded or not, I haven't really peered into the internals of skimage.

Yeah, that's fair, and totally fine when you're working around the issue in your application. For hooks that we supply, though, I prefer informed decisions, which sometimes require digging into packages' internals.

> reason 2) if the hook targets just the ones that are currently lazy_loaded, I imagine that with each version, as skimage updates more submodules, they will convert more and more to using this new way. So the hook will have to be updated in lockstep. I don't actually know what their roadmap is, I'm just being pessimistic about it. 

True, and this is a perennial problem. Not just with lazy-loading, but also with cythonized extensions (from which we cannot infer imports). 

> reason 3) I didn't find it added much to the overall build time in pyinstaller. If there are other reasons than wasted time to avoid this kind of overkill, let me know.

The primary reason is not so much the build time (after all, a single brute-force collect might have less overhead than checking for versions and performing fine-grained collection in individual hooks), but rather the build size. Which covers both unnecessary modules as well as unnecessary data files. 

> that being said, if you'd still prefer I pare it down to just the imports you mention, I can probably do that without a ton more effort.

I'll prepare my version of PR with fine-grained changes to supersede this one, but I'd ask to you test it with your application, to ensure that I haven't missed anything (that's not covered by the basic import tests).

ryandeon:
I just tested your changes, and they worked in my project. Thanks for taking this on!

PR #566, opened 2023-03-31
hooks: update scikit-image hooks for compatibility with 0.19.x and 0.20.0

rokm:
Update scikit-image hooks for compatibility with 0.19.x and 0.20.0. The 0.19.x series introduced custom lazy module loader, which in our case affects `skimage.filters`. In 0.20.0, they switched to `lazy_loader`, which requires us to collect `__init__.pyi` files. In addition `skimage.morphology` introduced two mandatory .npy data files that it tries to load at import.

Supersedes and closes #565.


rokm:
Oneshot test: https://github.com/rokm/pyinstaller-hooks-contrib/actions/runs/4577015129

(@bwoodsend any plans for merging your `oneshot` branch to the repo?)

bwoodsend (approved):

PR #567, opened 2023-03-31
ci: conditionally enable slow tests for scikit-image and scikit-learn

rokm:
Add steps that run slow tests for scikit-image and/or scikit-learn when the corresponding package versions are explicitly updated in `requirements-tests-libraries`.


rokm:
This will hopefully let us keep up with updates to those packages, and breakages caused by proliferation of lazy loading and/or module cythonization.

bwoodsend (approved):

PR #573, opened 2023-04-23
Add hook for pylsl

robertoostenveld:
This adds a hook for pylsl, i.e., the Python wrapper around labstreaminglayer,  which is also known as LSL.

The pylsl package loads the C++ liblsl library using CDLL. The logic for locating liblsl is replicated here; subsequently it is added to the binaries and packaged in the compiled version in `pylsl/lib` which ensures that  pylsl will be able to find it.

See also
-  https://labstreaminglayer.readthedocs.io
-  https://github.com/labstreaminglayer/pylsl
-  https://github.com/sccn/liblsl

I tested it successfully on macOS and Windows, using a few methods to install liblsl (using conda, using homebrew, and using a local copy and the PYLSL_LIB variable).

### DIFF
--- a/.github/workflows/pr-test.yml
+++ b/.github/workflows/pr-test.yml
@@ -99,7 +99,7 @@ jobs:
           # Install libdiscid (dependency of discid python package).
           brew install libdiscid
           # Install lsl library for pylsl
-          brew install lsl
+          brew install brew install labstreaminglayer/tap/lsl
 
       - name: Install dependencies
         shell: bash
--- a/.github/workflows/pr-test.yml
+++ b/.github/workflows/pr-test.yml
@@ -98,6 +98,8 @@ jobs:
           brew install pango
           # Install libdiscid (dependency of discid python package).
           brew install libdiscid
+          # Install lsl library for pylsl
+          brew install lsl
 
       - name: Install dependencies
         shell: bash
--- a/.github/workflows/pr-test.yml
+++ b/.github/workflows/pr-test.yml
@@ -99,7 +99,7 @@ jobs:
           # Install libdiscid (dependency of discid python package).
           brew install libdiscid
           # Install lsl library for pylsl
-          brew install brew install labstreaminglayer/tap/lsl
+          brew install labstreaminglayer/tap/lsl
 
       - name: Install dependencies
         shell: bash
--- a/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-pylsl.py
+++ b/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-pylsl.py
@@ -75,7 +75,3 @@ if libfile:
 else:
     logger.warning("liblsl shared library not found - pylsl will likely fail to work!")
     binaries = []
-
-
-if __name__ == '__main__':
-    print(binaries)
--- a/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-pylsl.py
+++ b/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-pylsl.py
@@ -1,72 +1,29 @@
+# ------------------------------------------------------------------
+# Copyright (c) 2023 PyInstaller Development Team.
+#
+# This file is distributed under the terms of the GNU General Public
+# License (version 2.0 or later).
+#
+# The full license is available in LICENSE.GPL.txt, distributed with
+# this software.
+#
+# SPDX-License-Identifier: GPL-2.0-or-later
+# ------------------------------------------------------------------
+
 import os
-import platform
-import struct
-import pathlib
-from ctypes import util
-from PyInstaller.utils.hooks import get_module_file_attribute, logger
+from PyInstaller.utils.hooks import logger
 
-"""
-Hook for pylsl, i.e., the Python wrapper around labstreaminglayer (aka LSL).
-The pylsl package loads the C++ liblsl library using CDLL. The logic for
-locating liblsl is replicated here; subsequently it is added to the binaries
-and packaged in the compiled version such that pylsl will be able to find it
-there as well.
-
-See also
--  https://labstreaminglayer.readthedocs.io
--  https://github.com/labstreaminglayer/pylsl
--  https://github.com/sccn/liblsl
-"""
-
-
-def find_liblsl_library():
-    """finds the binary lsl library.
-
-    Search order is to first try to use the path stored in the environment
-    variable PYLSL_LIB (if available), then search through the package
-    directory, and finally search the whole system.
-    """
-
-    # first try the environment variable PYLSL_LIB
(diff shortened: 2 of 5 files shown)


robertoostenveld:
I am not sure how I should implement the test. The reason pylsl needs a special hook is because it loads a library (`liblsl.so`, `liblsl.dylib` or `lsl.dll`). The library is not installed by default. The documentation at https://pypi.org/project/pylsl/

> On all non-Windows platforms and for some Windows-Python combinations, you must first obtain a liblsl shared library:
> ...

This requires either conda, homebrew, a "random" downloaded library, or compiling it yourself from the C++ source code. I don't think that any of these would work with the pyinstaller hooks test script.

src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-pylsl.py
```diff
@@ -0,0 +1,77 @@
...
+there as well.
+
+See also
+-  https://labstreaminglayer.readthedocs.io
+-  https://github.com/labstreaminglayer/pylsl
+-  https://github.com/sccn/liblsl
+"""
+
+
+def find_liblsl_library():
```
bwoodsend, line 22: Any reason why you can't just import and call [the original find_liblsl_library()](https://github.com/labstreaminglayer/pylsl/blob/16a4198087936386e866d7239bfde32d1fef6d6b/pylsl/pylsl.py#L1198)?

bwoodsend (commented):
> This requires either conda, homebrew, a "random" downloaded library, or compiling it yourself from the C++ source code. I don't think that any of these would work with the pyinstaller hooks test script.

If it's on brew, then installing it [here](https://github.com/pyinstaller/pyinstaller-hooks-contrib/blob/79f28de9dc7662dd011d19ed6f1d1db3ca006afe/.github/workflows/pr-test.yml#L95-L100) then using `; sys_platform == "darwin"` in the `requirements-test-libraries.txt` file would at least give us coverage on one platform which is a lot better than none at all.

src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-pylsl.py
robertoostenveld: Hmm, that is something I had not considered. That could indeed work, I will give it a try.

robertoostenveld: Yes, that works. The generator function itself is not directly exposed, but it can be imported with some extra effort.

I also added the test as you suggested.

rokm: I think it might actually be best to just read the  `pylsl.pylsl.libpath` attribute:

https://github.com/labstreaminglayer/pylsl/blob/16a4198087936386e866d7239bfde32d1fef6d6b/pylsl/pylsl.py#L1286-L1302

i.e.,

```python
from PyInstaller.utils.hooks import get_module_attribute

try:
    libpath = get_module_attribute('pylsl.pylsl', 'libpath')
except Exception:
    libpath = None
``` 

---

The test function (which should be added to https://github.com/pyinstaller/pyinstaller-hooks-contrib/blob/master/src/_pyinstaller_hooks_contrib/tests/test_libraries.py) should also look at `pylsl.pylsl.libpath` and ensure that the library is loaded from the frozen application's directory (i.e., that the path begin withs `sys._MEIPASS`). Otherwise, the test might still pass even though the library is not properly collected (because the system copy can still be found).

src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-pylsl.py
```diff
@@ -1,72 +1,29 @@
...
-                        if libfile and os.path.isfile(libfile):
-                            return libfile
-                    elif (scope == "system") and os_name == "Darwin":
-                        # try on the homebrew library path
-                        libfile = util.find_library('/opt/homebrew/lib/' + libprefix + "lsl" + bitness + debugsuffix)
-                        if libfile and os.path.isfile(libfile):
-                            return libfile
+try:
+    # the import will fail it the library cannot be found
+    from pylsl import pylsl
```
bwoodsend, line 18: Sorry, I should have said: Whenever a hook needs to load a 3rd party library it needs to be done in an isolated subprocess. See [here](https://pyinstaller.org/en/v5.10.1/hooks.html#module-PyInstaller.isolated). Basically put this in a function then use [isolated.call()](https://pyinstaller.org/en/v5.10.1/hooks.html#PyInstaller.isolated.call) on it.

robertoostenveld: I have now implemented this.

.github/workflows/pr-test.yml
```diff
@@ -98,6 +98,8 @@ jobs:
           brew install pango
           # Install libdiscid (dependency of discid python package).
           brew install libdiscid
+          # Install lsl library for pylsl
+          brew install lsl
```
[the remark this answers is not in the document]
robertoostenveld: somehow brew on my computer knew where to get it from, probably I told it in the past. It is indeed `labstreaminglayer/tap/lsl` 
```
macbook> brew uninstall lsl
Uninstalling /usr/local/Cellar/lsl/1.16.1... (23 files, 1.1MB)
macbook> brew install lsl
==> Fetching labstreaminglayer/tap/lsl
...
```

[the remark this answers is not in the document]
bwoodsend: You forgot to update in `.github/workflows/pr-test.yml`.

.github/workflows/pr-test.yml
```diff
@@ -98,6 +98,8 @@ jobs:
           brew install pango
           # Install libdiscid (dependency of discid python package).
           brew install libdiscid
+          # Install lsl library for pylsl
+          brew install brew install labstreaminglayer/tap/lsl
```
bwoodsend, line 102:
```suggestion
          brew install labstreaminglayer/tap/lsl
```

bwoodsend (changes requested):
And we'll need a test for pylsl too then.

src/_pyinstaller_hooks_contrib/tests/test_libraries.py
```diff
@@ -251,6 +251,15 @@ def test_markdown(pyi_builder):
         """)
 
 
+@importorskip('pylsl')
+def test_lxml_isoschematron(pyi_builder):
```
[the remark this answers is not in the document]
robertoostenveld: oh sorry,  that is a stupid copy-and-paste error. I'll fix it.

bwoodsend (approved):

bwoodsend:
Thanks @robertoostenveld

PR #575, opened 2023-04-26
Remove py hidden imports for pyshark >= 0.6.

bwoodsend:
`pyshark` now uses `termcolor` instead of `py._io.terminalwriter` for writing colored text to terminals.


bwoodsend:
Here's a oneshot run of the previous version of pyshark (0.5.3): https://github.com/bwoodsend/pyinstaller-hooks-contrib/actions/runs/4813393459

rokm (approved):

PR #579, opened 2023-05-11
Add hook for graphql_query

M0Rf30:
graphql_query needs its templates folder, otherwise the jinja2 loader will fail

### DIFF
--- a/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-graphql_query.py
+++ b/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-graphql_query.py
@@ -1,0 +1,17 @@
+#-----------------------------------------------------------------------------
+# Copyright (c) 2005-2023, PyInstaller Development Team.
+#
+# Distributed under the terms of the GNU General Public License (version 2
+# or later) with exception for distributing the bootloader.
+#
+# The full license is in the file COPYING.txt, distributed with this software.
+#
+# SPDX-License-Identifier: (GPL-2.0-or-later WITH Bootloader-exception)
+#-----------------------------------------------------------------------------
+"""
+PyInstaller hook file for graphql_query. Tested with version 1.0.3.
+"""
+
+from PyInstaller.utils.hooks import collect_data_files
+
+datas = collect_data_files('graphql_query.templates')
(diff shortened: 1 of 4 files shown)


src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-graphql_query.py
```diff
@@ -0,0 +1,19 @@
...
+#
+# SPDX-License-Identifier: (GPL-2.0-or-later WITH Bootloader-exception)
+#-----------------------------------------------------------------------------
+"""
+PyInstaller hook file for graphql_query. Tested with version 1.0.3.
+"""
+
+from PyInstaller.utils.hooks import collect_data_files
+
+hiddenimports = ["graphql_query"]
```
rokm, line 17: In order for the hook to be ran, the module already had to be discovered. So this hidden import is definitely not required.

rokm:
Can you also add a basic test that demonstrates the need for the template files (assuming the basic example from https://github.com/denisart/graphql-query will do?) to [test_libraries.py](https://github.com/pyinstaller/pyinstaller-hooks-contrib/blob/master/src/_pyinstaller_hooks_contrib/tests/test_libraries.py) and then add `graphql-query==1.0.3` to [requirements-test-libraries.txt](https://github.com/pyinstaller/pyinstaller-hooks-contrib/blob/master/requirements-test-libraries.txt) - this will also trigger the CI on all OSes. 

Then [add a news fragment](https://github.com/pyinstaller/pyinstaller-hooks-contrib/tree/master/news), and we're good to go.

src/_pyinstaller_hooks_contrib/tests/test_libraries.py
```diff
@@ -1214,6 +1214,14 @@ def test_gitlab(pyi_builder):
         import gitlab
         """)
```
bwoodsend, line 1216: The linter wants an extra blank line here.

src/_pyinstaller_hooks_contrib/tests/test_libraries.py
```diff
@@ -1214,6 +1214,14 @@ def test_gitlab(pyi_builder):
...
 
+@importorskip('graphql_query')
+def test_graphql_query(pyi_builder):
+    pyi_builder.test_source("""
+        from graphql_query import Operation, Query
+        hero = Query(name="hero", fields=["name"])
+        operation = Operation(type="query", queries=[hero])
+        print(operation.render())
+        """)
```
bwoodsend, line 1225: ... and here.

bwoodsend:
That's it released in `pyinstaller-hooks-contrib==2023.3`.

PR #582, opened 2023-05-13
Add hook for librosa

rokm:


src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-librosa.py
```diff
@@ -0,0 +1,22 @@
...
+# SPDX-License-Identifier: GPL-2.0-or-later
+# ------------------------------------------------------------------
+
+from PyInstaller.utils.hooks import collect_data_files, collect_submodules
+
+# Collect all data files from the package. These include:
+#  - package's and subpackages' .pyi files for `lazy_loader`
+#  - example data in librosa/util, required by `librosa.util.files`
+#  - librosa/core/intervals.msgpack, required by `librosa.core.intervals`
+datas = collect_data_files("librosa", excludes=['**/__pycache__'])
```
bwoodsend, line 19: Why does this need `excludes=['**/__pycache__']`? Pycache should be excluded by default?

rokm: I could swear I saw it collected yesterday...

rokm: So I did, but it was in an environment where I have already run some `librosa.util` functions. That triggered `numba` compilation, which placed its files into `__pycache__`, e.g.:
- `venv/lib/python3.11/site-packages/librosa/util/__pycache__/guf-utils._localmin-1080.py311.1.nbc`
- `venv/lib/python3.11/site-packages/librosa/util/__pycache__/guf-utils._localmin-1080.py311.nbi`

And while .pyc files from `__pycache__` are not picked up by `collect_data_files`, these `.nbi` and `.nbc` files are. 

However, they are dead-weight, because the frozen application does not seem to re-use them, and compiles new set using global cache directory (e.g., `~/.cache/numba`). This is because the unfrozen numba caching codepath uses `_InTreeCacheLocator`, while the frozen one uses `_UserWideCacheLocator` - presumably because the `source_path` in [this check](https://github.com/numba/numba/blob/e4ff3cf5fcb59e91fc7d46c340f2e7eb664dd0c0/numba/core/caching.py#L331-L339) does not point to a valid file...

I'll update the comment to reflect this.

bwoodsend (approved):

PR #588, opened 2023-05-23
hooks: update bokeh hook for bokeh >= 3.0.0

rokm:
Starting with `bokeh` 3.0.0, the package sets its `__version__` from metadata, so we need to collect it.

This one flew under our radar because although we have a hook for `bokeh`, we do not have an active test. And the `panel` test failing in the yesterday's weekly update PR is probably due to `panel` re-pinning required version of `bokeh`.


bwoodsend:
I've been looking at why this one wasn't detected automatically by PyInstaller's bytecode scanning. Turns out that `import importlib.metadata as importlib_metadata` breaks it because [_metadata_from()](https://github.com/pyinstaller/pyinstaller/blob/7bc13367912318f1887e714d33c24338b9f84afb/PyInstaller/depend/analysis.py#L787) only searches for `importlib.metadata.version` when scanning modules importing `importlib.metadata` and for `importlib_metadata.version` when scanning modules importing `importlib_metadata`. It would have needed to search for `importlib_metadata.version` when a module imports `importlib.metadata`. I guess this is probably more fiddly to fix in PyInstaller.

bwoodsend (approved):

rokm:
I always forget that metadata should be picked up automatically...

One general limitation of the bytecode scanning is that we have no deal of tracking aliases back to their original names. In addition to `importlib.metadata` being imported as `importlib_metadata`, we currently also have no means of tracking `version` back to `from importlib.metadata import version` - which is annoying, because `version` could hypothetically as well be a local function with single string argument, so having `_metadata_from` pick up on just `version` might result in false positives...

PR #595, opened 2023-06-06
test requirements: restrict adbutils to python < 3.10

rokm:
The latest release of `adbutils` (1.2.11) explicitly pins `requests<2.16`, resulting in installation of `requests-2.15.1`, which is incompatible with python >= 3.10 due to use of `collections.Mapping` that was moved to `collections.abc`.

Hence, the package is essentially incompatible with python >= 3.10, and we should not be testing it there.


bwoodsend (approved):
I can't even tell if >3.8 is supposed to be supported or not from looking at that repo.

PR #611, opened 2023-07-04
hooks: update pydantic hook for compatibility with pydantic v2.0.0

### LINKED ISSUE #610: Pydantic.compiled Depreciated
Got this error when compiling something with Pydantic in it:

`pydantic.errors.PydanticImportError: `pydantic:compiled` has been removed in V2.`

From:
https://github.com/pyinstaller/pyinstaller-hooks-contrib/blob/master/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-pydantic.py

rokm:
The v2.0.0 of pydantic seems to have removed the `pydantic.compiled` attribute, and cython-compiled version of PyPI wheels do not seem to be available anymore.

Fixes #610.

### DIFF
--- a/news/611.update.rst
+++ b/news/611.update.rst
@@ -1,0 +1,1 @@
+Update ``pydantic`` hook for compatibility with ``pydantic`` v2.0.0.
--- a/requirements-test-libraries.txt
+++ b/requirements-test-libraries.txt
@@ -133,6 +133,7 @@ librosa==0.10.0.post2
 sympy==1.12
 xyzservices==2023.5.0
 mistune==3.0.1
+pydantic==2.0.0
 
 # ------------------- Platform (OS) specifics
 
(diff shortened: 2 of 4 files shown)


bwoodsend (approved):

bwoodsend:
Glad one library has finally realised that blindly cythonizing everything is more trouble than it's worth.

Wyko:
Nice fast response to this issue!

alekssamos:
Hi. Everything has been updated to the latest versions, but it still doesn't work.

rokm:
> Hi. Everything has been updated to the latest versions, but it still doesn't work.

What is the pydantic version you are using, and what is the error traceback?

alekssamos:
Strange. What's it? It's working. Somehow it worked by itself. And it didn't work 10 minutes ago. I'm sorry, it's okay.
I think I've run the command `pip install -U pyinstaller-hooks-contrib`
And it was written that everything has already been installed and so the newest and no changes have been made.
And after that, I thought that nothing had changed.
And I wrote here, because there was an error before executing this command.
I am surprised by the instant response! Thank you very much! All the best!

PR #617, opened 2023-07-19
hooks: add hook for ens package

rokm:
Add hook for `ens` package to collect its data files. Required by `web3` >= 6.6.0.

### DIFF
--- a/requirements-test-libraries.txt
+++ b/requirements-test-libraries.txt
@@ -99,7 +99,7 @@ Twisted==22.10.0
 tzdata==2023.3
 Unidecode==1.3.6
 weasyprint==59.0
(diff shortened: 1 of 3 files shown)


requirements-test-libraries.txt
```diff
@@ -99,7 +99,7 @@ Twisted==22.10.0
 tzdata==2023.3
 Unidecode==1.3.6
 weasyprint==59.0
-web3==6.5.0
+web3==6.6.1
 websockets==11.0.3
```
bwoodsend, line 103: Should we not also pin `ens`?

bwoodsend (approved):

PR #621, opened 2023-07-31
hooks: update graphql_query hook for compatibility with v1.2.0

rokm:
Update `graphql_query` hook for compatibility with `graphql-query` v1.2.0. Collect data files from `graphql_query` instead of `graphql_query.templates`; the templates directory cannot be treated as a namespace subpackage anymore due to introduction of eponymous submodule.


bwoodsend (approved):

rokm:
Fixing an incorrect PR number in news fragment...

PR #626, opened 2023-08-19
hooks: add hook for Lingua

BLKSerene:


src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-lingua-language-detector.py
bwoodsend, line 1: This file should be named after importable package rather than the pip installable distribution. i.e. `hook-lingua.py` rather than `hook-lingua-language-detector.py`.

src/_pyinstaller_hooks_contrib/tests/test_libraries.py
```diff
@@ -1723,3 +1723,10 @@ def sync_hello_world() -> Dict[str, Any]:
         assert response.status_code == 200
         assert response.json() == {"hello": "world"}
     """)
+
+
+@importorskip('lingua')
+def test_lingua_language_detector(pyi_builder):
+    pyi_builder.test_source("""
+        import lingua
```
bwoodsend, line 1731: This test passes even without the hook. To make things more awkward, it looks like `lingua` doesn't raise any errors when its data files are missing – rather it just silently returns 0 probabilities and `None` detected languages. Perhaps you could use [this example](https://pemistahl.github.io/lingua-py/lingua.html#71-basic-usage) with an 

```python
assert detector.detect_language_of("languages are awesome") == Language.ENGLISH
```
at the end?

bwoodsend (changes requested):

bwoodsend (approved):

bwoodsend:
Nice work @BLKSerene

PR #628, opened 2023-08-19
hooks: add hook for jieba

BLKSerene:
This PR adds a hook for [jieba](https://github.com/fxsjy/jieba) which is a Chinese word tokenization library.

### DIFF
--- a/news/628.new.rst
+++ b/news/628.new.rst
@@ -1,0 +1,1 @@
+Add hook for ``jieba``.
--- a/requirements-test-libraries.txt
+++ b/requirements-test-libraries.txt
@@ -139,6 +139,7 @@ psutil==5.9.5
 litestar==2.0.0rc1
 lingua-language-detector==1.3.2; python_version >= '3.8'
 opencc-python-reimplemented==0.1.7
+jieba==0.42.1
 
 # ------------------- Platform (OS) specifics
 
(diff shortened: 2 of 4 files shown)


bwoodsend:
Hmm, the pull request is good but I'm rather uncomfortable with the library itself. I just opened up a file randomly to see what it does with its data files and I see [this](https://github.com/fxsjy/jieba/blob/67fa2e36e72f69d9134b8a1037b83fbb070b9775/jieba/_compat.py#L23-L24) accessing data files from the wrong place, [this](https://github.com/fxsjy/jieba/blob/67fa2e36e72f69d9134b8a1037b83fbb070b9775/jieba/_compat.py#L32) trying to install a package on the fly into potentially the wrong Python environment (which wouldn't work with PyInstaller) and [a version comparison](https://github.com/fxsjy/jieba/blob/67fa2e36e72f69d9134b8a1037b83fbb070b9775/jieba/_compat.py#L39) which will break when `paddle 1.10.0` is released.

I'm happy to accept the pull request but I want you to be aware that I'm more likely to just mark the test as skipped and ignore this library than fix it at the first sign of `jieba` induced trouble.

BLKSerene:
The issue of installing `paddlepaddle` and the version comparison issue would occur when `enable_paddle` is called to use the paddle mode of `jieba`. For my use cases, it is fine.

There have been no updates for the library for several years, and I don't think they would have the time to review PRs even if submit one. So, shall I just remove the test?

bwoodsend:
If they're not publishing releases then that test will probably never run again (at least not automatically) so it doesn't make much difference whether or not it stays. @rokm What do you reckon? I say we just merge as is?

rokm:
Yes, let's merge as-is. We'll deal with any trouble rising from updates if/when they actually happen.

PR #629, opened 2023-08-19
hooks: add hook for simplemma

BLKSerene:
This PR adds a hook for [simplemma](https://github.com/adbar/simplemma), which is a simple multilingual lemmatizer.

### DIFF
--- a/src/_pyinstaller_hooks_contrib/tests/test_libraries.py
+++ b/src/_pyinstaller_hooks_contrib/tests/test_libraries.py
@@ -1755,3 +1755,12 @@ def test_jieba(pyi_builder):
 
         assert jieba.lcut('我来到北京清华大学') == ['我', '来到', '北京', '清华大学']
     """)
+
+
+@importorskip('simplemma')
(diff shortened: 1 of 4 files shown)


src/_pyinstaller_hooks_contrib/tests/test_libraries.py
```diff
@@ -1755,3 +1755,12 @@ def test_jieba(pyi_builder):
...
         assert jieba.lcut('我来到北京清华大学') == ['我', '来到', '北京', '清华大学']
     """)
+
+
+@importorskip('simplemma')
+def test_simplemma(pyi_builder):
+    pyi_builder.test_source("""
+        import simplemma
+
+        assert simplemma.lemmatize('tests', lang = 'en') == 'test'
```
rokm, line 1765: The linter won't complain because the source is in the string, but those spaces in `lang = 'en'` would cause `E251 unexpected spaces around keyword / parameter equals`.

rokm (approved):

rokm:
Thanks again!

PR #630, opened 2023-08-19
hooks: add hook for wordcloud

BLKSerene:
This PR adds a hook for [wordcloud](https://github.com/amueller/word_cloud), which is a word cloud generator.

### DIFF
--- a/news/630.new.rst
+++ b/news/630.new.rst
@@ -1,0 +1,1 @@
+Add hook for ``wordcloud``.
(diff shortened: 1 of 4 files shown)


bwoodsend (approved):

bwoodsend:
I don't suppose you're familiar with git branching? It would mean that you don't have to submit all these hooks one at a time.

BLKSerene:
@bwoodsend Well, I think that submit one hook at a time (for packages with quite different functionalities) would make the commit history clearer and easier to look into if something wrong happens in the future.

bwoodsend:
No, branching solves the problem of commit history. Each new hook would still be a separate commit.

PR #632, opened 2023-08-22
hooks: add hook for python-mecab-ko

BLKSerene:
This PR adds a hook for [python-mecab-ko](https://github.com/jonghwanhyeon/python-mecab-ko), which is a python binding for [mecab-ko](https://bitbucket.org/eunjeon/mecab-ko).

### DIFF
--- a/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-mecab.py
+++ b/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-mecab.py
@@ -1,0 +1,16 @@
+# ------------------------------------------------------------------
+# Copyright (c) 2023 PyInstaller Development Team.
+#
+# This file is distributed under the terms of the GNU General Public
+# License (version 2.0 or later).
+#
+# The full license is available in LICENSE.GPL.txt, distributed with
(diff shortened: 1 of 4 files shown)


src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-mecab.py
```diff
@@ -0,0 +1,16 @@
...
+# The full license is available in LICENSE.GPL.txt, distributed with
+# this software.
+#
+# SPDX-License-Identifier: GPL-2.0-or-later
+# ------------------------------------------------------------------
+
+from PyInstaller.utils.hooks import collect_data_files
+
+datas = collect_data_files('mecab')
+datas = collect_data_files('mecab_ko_dic')
```
rokm, line 16: This should be `datas += collect_data_files('mecab_ko_dic')`, otherwise you lose the `mecabrc` file from `mecab` package that you collected in previous call.

rokm:
I've pushed an update for MariaDB setup script hash, so if you rebase your open PRs on top of `master`, the linux tests should also work now.

PR #633, opened 2023-08-22
hooks: add hook for khmer-nltk and python-crfsuite

BLKSerene:
This PR adds a hook for [khmer-nltk](https://github.com/VietHoang1512/khmer-nltk), which is a Khmer (Cambodian) natural language processing toolkit.

### DIFF
--- a/news/633.new.1.rst
+++ b/news/633.new.1.rst
@@ -1,0 +1,1 @@
+Add hook for ``python-crfsuite``.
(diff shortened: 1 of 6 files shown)


rokm:
Looks like this will require a hook for `pycrfsuite`.

BLKSerene:
@rokm Shall I add another separate hook for `pycrfsuite` or just add it here?

rokm:
It should go in a separate hook, because this error will be encountered by anyone trying to import `pycrfsuite`. But you can add the hook in this PR, in a separate commit (and you can add additional news fragment called `633.news.1.rst` for that change.

bwoodsend (approved):

PR #634, opened 2023-08-22
hooks: add hook for pymorphy3

BLKSerene:
This PR adds a hook for [pymorphy3](https://github.com/no-plagiarism/pymorphy3), which is a fork of [pymorphy2](https://github.com/pymorphy2/pymorphy2), which then is a morphological analyzer for Russian and Ukrainian.

The Ukrainian model (`pymorphy3-dicts-uk`) is not installed automatically as a dependency of `pymorphy3`, so it needs to be handled separately.

Notes: The maintenance of the [pymorphy2](https://github.com/pymorphy2/pymorphy2) project has stalled (pymorphy2/pymorphy2#160), and [pymorphy3](https://github.com/no-plagiarism/pymorphy3) seems to be the most active fork available now (the [spaCy](https://spacy.io/) project widely used for multilingual NLP has already switched from `pymorphy2` to `pymorphy3` in explosion/spaCy#11345).

### DIFF
--- a/news/634.new.rst
+++ b/news/634.new.rst
@@ -1,0 +1,1 @@
+Add hook for ``pymorphy3``.
--- a/requirements-test-libraries.txt
+++ b/requirements-test-libraries.txt
@@ -146,6 +146,8 @@ eng-to-ipa==0.0.2
 python-mecab-ko==1.3.3
 khmer-nltk==1.5
 python-crfsuite==0.9.9
+pymorphy3==1.2.0
+pymorphy3-dicts-uk==2.4.1.1.1663094765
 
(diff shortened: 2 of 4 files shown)


rokm:
You can use `import pymorphy3_dicts_uk  # noqa: F401` to work around the linter.

Although, we generally try to avoid directly importing the packages in the hook (and hence in the main PyInstaller process). You can use `PyInstaller.utils.hooks.can_import_module` helper instead (which tries to import the module/package in an isolated subprocess).

BLKSerene:
flake8 should be happy now.

rokm (approved):

PR #636, opened 2023-08-23
Fix an issue with enchant hook

rsnyman:
Fix an issue with enchant 2 using a different directory (in MacPorts)

### DIFF
--- a/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-enchant.py
+++ b/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-enchant.py
@@ -59,4 +59,7 @@ if is_darwin:
         # In Macports are available mostly hunspell (myspell) and aspell dictionaries.
         libdir = os.path.dirname(libenchant)  # e.g. /opt/local/lib
         sharedir = os.path.join(os.path.dirname(libdir), 'share')  # e.g. /opt/local/share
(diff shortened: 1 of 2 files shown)


src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-enchant.py
```diff
@@ -59,4 +59,7 @@
         # In Macports are available mostly hunspell (myspell) and aspell dictionaries.
         libdir = os.path.dirname(libenchant)  # e.g. /opt/local/lib
         sharedir = os.path.join(os.path.dirname(libdir), 'share')  # e.g. /opt/local/share
-        datas.append((os.path.join(sharedir, 'enchant'), 'enchant/share/enchant'))
+        if os.path.exists(os.path.join(sharedir, 'enchant')):
+            datas.append((os.path.join(sharedir, 'enchant'), 'enchant/share/enchant'))
+        if os.path.exists(os.path.join(sharedir, 'enchant-2')):
+            datas.append((os.path.join(sharedir, 'enchant-2'), 'enchant/share/enchant'))
```
bwoodsend, line 65: The destination should probably `enchant/share/enchant-2' so that if both exist, they don't clobber each other?

rsnyman: Good point. I just tested it out, and it appears my app is able to find the dictionaries without an issue.

bwoodsend (approved):

PR #637, opened 2023-08-28
Add license to PyPI classifiers

WilliamRoyNelson:
Adds license to [PyPI project classifiers](https://pypi.org/classifiers/)

The reason that it is very important for this information to be present is that in an enterprise environment, security tools like [Sonatype Nexus IQ](https://help.sonatype.com/iqserver) are used to manage open source software risk. Nexus IQ specifically can be configured to [classify packages according to their license](https://help.sonatype.com/iqserver/managing/policy-management/license-threat-groups). This allows companies to control which licenses are used internally to prevent accidental use of an overly restrictive license.

If no license is included in the PyPI classifiers, then automated tools may assume that no license exists and the software cannot be used:
image

Adding this license information will increase the availability of PyInstaller within enterprise environments.

I'm not 100% sure what the most appropriate license to use is. I just used the license classifier in PyInstaller: https://github.com/pyinstaller/pyinstaller/blob/develop/setup.cfg#L29

If a different choice is more appropriate, I can change it.


bwoodsend:
Ughh, another blunt tool!

I'm not sure a trove classifier can be applied to something that's non-uniformly licensed. Most of the code is GPL2, some of it is MIT as well as GPL2, the Python code that PyInstaller will embed in your application is APL and then there's a waiver saying that the bootloaders are still GPL but can be included in any application regardless of license. 

In practice though, this shouldn't matter since PyInstaller is a build tool rather than a runtime dependency. Our license does not apply to your application, nor are you required to distribute our license in anything you build. Can you really not just add a *not applicable* exception to this Nexus IQ thing?

rokm:
I suppose in the grand scheme of things, it shouldn't hurt if this package's License Classifier matched the one from the main PyInstaller package? Even if neither really covers the run-time hooks, which are the part that actually matters for end-users...

WilliamRoyNelson:
> In practice though, this shouldn't matter since PyInstaller is a build tool rather than a runtime dependency. Our license does not apply to your application, nor are you required to distribute our license in anything you build. Can you really not just add a _not applicable_ exception to this Nexus IQ thing?

I completely understand where you're coming from. You are correct that tools like Nexus IQ do have options to exclude items from classification. The issue is that depending on the size/sophistication of the organization, getting that exclusion could be as simple as walking down the hall to ask a friend to add it to the list, or as complicated as sending the whole thing through legal review.

The unfortunate part about security tools like this is that from the perspective of a less experienced Python user, the problem is that they typed `pip install pyinstaller` and they get an error, so they conclude that PyInstaller is blocked at work.

As a side note, while we're looking at the classifiers, is there anything else that ought to get changed? Now's as good a time as any!

bwoodsend:
> The issue is that depending on the size/sophistication of the organization, getting that exclusion could be as simple as walking down the hall to ask a friend to add it to the list, or as complicated as sending the whole thing through legal review.

Hmm, fair enough. Makes me think I've maybe been a little too cutthroat in not bumping PyInstaller's zlib version whenever people's silly little BlackDuck scanners start moaning about non-applicable vulnerabilities.

I'm still not really sure what the right answer is here but I think adding `License :: OSI Approved :: Apache Software License` as well as the GPL2 classifier makes the most sense to me. Happy to be overruled here.

rokm:
> I'm still not really sure what the right answer is here but I think adding `License :: OSI Approved :: Apache Software License` as well as the GPL2 classifier makes the most sense to me. Happy to be overruled here.

Sounds reasonable to me. Both here and in the main PyInstaller repository.

WilliamRoyNelson:
> > The issue is that depending on the size/sophistication of the organization, getting that exclusion could be as simple as walking down the hall to ask a friend to add it to the list, or as complicated as sending the whole thing through legal review.
> 
> Hmm, fair enough. Makes me think I've maybe been a little too cutthroat in not bumping PyInstaller's zlib version whenever people's silly little BlackDuck scanners start moaning about non-applicable vulnerabilities.
> 
> I'm still not really sure what the right answer is here but I think adding `License :: OSI Approved :: Apache Software License` as well as the GPL2 classifier makes the most sense to me. Happy to be overruled here.

Off the top of my head, I don't know of any packages that have multiple licenses in the classifiers so I tried it out with a test package just to make sure it doesn't make PyPI explode:
https://test.pypi.org/project/example-package-WilliamRoyNelson/0.0.1/
image

Looks fine to me! I've already pushed the suggested change for the Apache Software License

bwoodsend (approved):

rokm (approved):

PR #641, opened 2023-09-11
tests: allow test_pymorphy3 to run without the uk dictionary installed

rokm:
Modify the `pymorphy3` test to dynamically determine what dictionaries (`ru`, `uk`) are available in the environment, and test for that. Allows the test to run when the pyup pull request updates only the `morphy3` package (which installs only `ru` dictionary by default).


rokm:
A oneshot CI that simulates the situation in #640: https://github.com/rokm/pyinstaller-hooks-contrib/actions/runs/6151565650

PR #646, opened 2023-09-28
hooks: update runtime hook for NLTK

BLKSerene:
If `nltk_data` can be found both in the frozen program and under the default location specified by `NLTK`, the former should be preferred to the latter.

### DIFF
--- a/news/646.update.rst
+++ b/news/646.update.rst
@@ -1,0 +1,1 @@
+If ``nltk_data`` can be found both in the frozen program and under the default location specified by ``NLTK``, the former should be preferred to the latter.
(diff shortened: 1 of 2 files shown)


news/644.new.rst
```diff
@@ -0,0 +1 @@
+If `nltk_data` can be found both in the frozen program and under the default location specified by `NLTK`, the former should be preferred to the latter.
```
rokm, line 1: Use two backticks to quote nltk_data and NLTK. The file should be called `646.update.rst`.

rokm (approved):

PR #648, opened 2023-09-30
Drop Pywin32 stdhook

byehack:
they are already hardcoded in pyinstaller package.


rokm:
I'm pretty sure this breaks the following test-case programs:
```
import pywintypes
print("Hello", pywintypes)
```
and
```
import pythoncom
print("Hello", pythoncom)
```

I.e., when these two modules are not also picked up by the binary dependency analysis (which is the part that is hardcoded in PyInstaller).

byehack:
> when these two modules are not also picked up by the binary dependency analysis

Correct, I tested against importing other modules those didn't affect this.

Actually, I was [trying to disable](https://github.com/byehack/pyinstaller/commit/36b6d1434512b2e500665f82da6d2c8c007ba9e3) **preserve dll parent structure** feature, then these hooks prevent to disable it.

I use PyInstaller to pack requirement libraries that entry app imports.  I can't have access runtime-hooks, so I need to all dlls be in the root folder like "pyinstaller<5.4".

byehack:
So, now I think per https://github.com/pyinstaller/pyinstaller/commit/be260d46acdef7bd3c5b2c9ba4b424ab0b258a8a we can at least exclude `pyinstaller>=5.12"` for these hook.

rokm:
I'd rather the hook did the correct thing, independent of what core PyInstaller does for the binary analysis counterpart. So this change seems pointless to me.

PR #659, opened 2023-11-01
Create hook-coincurve

earonesty:
coincurve hooks, tested osx/linux/windows

### DIFF
--- a/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-coincurve
+++ b/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-coincurve
@@ -1,0 +1,26 @@
+from PyInstaller.utils.hooks import collect_data_files, get_package_paths
+import os
+import re
+
+hiddenimports = ['_cffi_backend']
+
+# Get the package path
+package_path = get_package_paths('coincurve')[0]
+
+# Collect data files
+datas = collect_data_files('coincurve')
+
+# Regular expression pattern for libsecp256k1
+libsecp256k1_pattern = re.compile(r'(libsecp256k1|_libsecp256k1).*\.(dll|so)')
+
+# Search for libsecp256k1 file within the 'coincurve' folder
+coincurve_path = os.path.join(package_path, 'coincurve')
+libsecp256k1_path = None
+for root, _, files in os.walk(coincurve_path):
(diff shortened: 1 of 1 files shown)


src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-coincurve
```diff
@@ -0,0 +1,38 @@
...
+import os
+import re
+
+hiddenimports = ['_cffi_backend']
+
+# Get the package path
+package_path = get_package_paths('coincurve')[0]
+
+# Collect data files
+datas = collect_data_files('coincurve')
```
rokm, line 23: What data files are there to collect?

src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-coincurve
```diff
@@ -0,0 +1,38 @@
...
+coincurve_path = os.path.join(package_path, 'coincurve')
+libsecp256k1_path = None
+for root, _, files in os.walk(coincurve_path):
+    for file in files:
+        if libsecp256k1_pattern.search(file):
+            libsecp256k1_path = os.path.join(root, file)
+            datas.append((libsecp256k1_path, 'coincurve'))
+            break
+
+    assert libsecp256k1_path, f"can't find coincurve libsecp256k1, look in {coincurve_path} and then modify this hook"
```
rokm, line 38: Any reason this is not using `binaries = collect_dynamic_libs('coincurve')`? This should pick up the DLL on Windows. 

On linux and macOS, `_libsecp256k1` seems to be an extension and should be picked up automatically.

rokm (changes requested):
Please also add a basic import test [here](https://github.com/pyinstaller/pyinstaller-hooks-contrib/blob/master/src/_pyinstaller_hooks_contrib/tests/test_libraries.py), and add `coincurve==18.0.0` to [requirements-test-libraries.txt](https://github.com/pyinstaller/pyinstaller-hooks-contrib/blob/master/requirements-test-libraries.txt) so that CI will run the added test.

earonesty:
I wasn't aware of that function should probably work just fine I would need
to test it

I've never submitted a hook just getting stuff to work that's all

PR #663, opened 2023-11-11
Pick up beta releases in check for pydantic 2.0.0+

monorimet:
Some pre-releases of pydantic have versions in the form "2.X.XbX" so update the conditional in the hook to recognize those versions as well.


rokm:
Thanks for bringing this to our attention! I see that `pydantic >= 2.0.0` indeed fails to match `2.5.0b1`.

This is regression introduced by removal of `pkg_resources` (#7943). The old `pkg_resources` based `is_module_satisfies` matched pre-release versions, while the new `is_module_satisfies` / `check_requirement` do not; but they should (by default) because currently all our hooks depend on that behavior.

PR #673, opened 2023-12-18
tests: fix dependencies of sudachipy test

rokm:
`sudachipy` test depends on availability of dictionary packages (`sudachidict_small`, `sudachidict_core`, and `sudachidict_full`), so skip the test if they are not available.

In practice, this means that we likely will not be running this test on the pyup-bot PRs, unless both `sudachipy` and dictionary packages get updated at the same time.


rokm:
Well, this was... interesting and fun. On the plus side, the SNAFU with `SudachiDict_full-0.0.0-py3-none-any.whl` did reveal that we need to add `sudachipy.errors` to hiddenimports for v0.6.8.

bwoodsend (approved):

Issue #562, opened 2023-03-22
PyInstaller issue with `transformers\__init__.py` file not found


Abhishek-Varma:
Not sure what's the right place for this issue that I'm facing.

I'm trying to use PyInstaller to make an executable for a python program.

Steps to reproduce the error mentioned in this thread :-
1: Fetch [myprog.py](https://gist.github.com/Abhishek-Varma/71ca5a5802570b8c293551d8cf6302d4).
2: Fetch the concerned [myprog_installer.spec](https://gist.github.com/Abhishek-Varma/71ca5a5802570b8c293551d8cf6302d4).
3: Download a [.safetensor file](https://huggingface.co/Linaqruf/anything-v3.0/resolve/main/anything-v3-fp16-pruned.safetensors).
4: Run `pyinstaller myprog_installer.spec`.
5: Run `.\myprog_cli.exe --checkpoint_path=/path/to/the/safetensors/file`.
     Eg: `.\myprog_cli.exe --checkpoint_path=C:\A\anything-v3-fp16-pruned.safetensors`
             
I am bumping into the following error :-
```
Traceback (most recent call last):
  File "transformers\utils\import_utils.py", line 1126, in _get_module
  File "importlib\__init__.py", line 126, in import_module
  File "<frozen importlib._bootstrap>", line 1206, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1178, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1149, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 690, in _load_unlocked
  File "PyInstaller\loader\pyimod02_importers.py", line 352, in exec_module
  File "transformers\models\blip\processing_blip.py", line 21, in <module>
  File "<frozen importlib._bootstrap>", line 1178, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1149, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 690, in _load_unlocked
  File "PyInstaller\loader\pyimod02_importers.py", line 352, in exec_module
  File "transformers\processing_utils.py", line 30, in <module>
  File "transformers\utils\import_utils.py", line 1155, in direct_transformers_import
  File "<frozen importlib._bootstrap_external>", line 936, in exec_module
  File "<frozen importlib._bootstrap_external>", line 1073, in get_code
  File "<frozen importlib._bootstrap_external>", line 1130, in get_data
[...]

PR #676, opened 2023-12-20
Add bunch of hooks for contemporary ML stuff

### LINKED ISSUE #462: Transformers hook needed
<!-- NO IMAGES --!>

transformers

Eg `foobar`

No

**Additional context**

Add any other context here.

rokm:
Add bunch of hooks for contemporary machine-learning stuff, along with updating the existing hooks for `torch` and `tensorflow`.

We now have hooks for `torchaudio` and `torchtext` to collect the dynamically-loaded extensions, and source .py files for TorchScript. So we can close https://github.com/pyinstaller/pyinstaller-hooks-contrib/issues/375, which was initially about missing extension libs (and the rest of the horrors in there should be resolved in PyInstaller 6.x). The `torchvision` hook is similarly updated to collect source .py files, and I've added a hook for `torchvision.io.image` to ensure that this module collects the `torchvision.image` extension that it dynamically loads.

The `tensorflow` hook got bunch of updates as well:
- with PyInstaller >= 6.0, we do not need to exclude `tensorflow.python._pywrap_tensorflow_internal` as a module anymore (in an attempt to avoid duplication); this exclusion was the cause of #121, so we can now close that.
- the version check if now done based on disttribution name (which we need to guess, due to different possibilities) - this addresses a potential regression from PyInstaller v5, because there `is_module_satisfies` implicitly fell back to checking `tensorflow.__version__`, while in v6, `is_module_satisfies` is purely metadata-based.
- we collect source files from `tensorflow.python.autograph` to avoid run-time warning about AutoGraph being unavailable
- we now collect plugins from `tensorflow-plugins`, and the run-time hook now monkey-patches `site.getsitepackages` to work around a faulty module file location check, which prevents the collected plugins from being loaded.

Both `torch` and `tensorflow` hooks now add hiddenimports to `nvidia.*` subpackages, which seem to be contemporary way of providing CUDA libs via PyPI wheels on Linux.

Then there's a hook for Hugging Face transformers - this one primarily deals with collection of metadata of all its dependencies; we collect metadata for all listed dependencies that are available at build time, to ensure they are also visible at run time (since we have no way of knowing what will be used). Plus, there's the source .py collection. This closes #462 and #562.

In addition, there are hooks for several other packages that we have seen in recent reports:
https://github.com/pyinstaller/pyinstaller/issues/5672
https://github.com/pyinstaller/pyinstaller/issues/5729
https://github.com/pyinstaller/pyinstaller/issues/7647
https://github.com/pyinstaller/pyinstaller/issues/7906
https://github.com/pyinstaller/pyinstaller/issues/7911
https://github.com/pyinstaller/pyinstaller/issues/7918
https://github.com/pyinstaller/pyinstaller/issues/7985
https://github.com/pyinstaller/pyinstaller/issues/8013
https://github.com/pyinstaller/pyinstaller/issues/8155

Primarily, these are about collection of source .py files for packages that use TorchScript or some other form of JIT. While the plan in https://github.com/pyinstaller/pyinstaller/issues/6290 is to automatically detect such cases, I think at this point it is more feasible to add hooks for the more popular packages.

There are bunch of tests added, but these are mostly there so that I could automatically test on different platforms (my main linux box, Windows laptop, and arm64 Mac) without having to juggle with test program scripts. They are definitely not meant for CI - both due to speed and because some of them need to download datasets and/or models. I did not put them under slow mark, but we can do so if we see them triggered by accident (i.e., something in our CI manages to pull in `torch` & co. as part of its dependencies).


rokm:
One caveat remains - on Linux, even if we collect CUDA libs from `nvidia.*` packages, `tensorflow` fails to discover them at run-time.

I think the reason is that `libtensorflow_cc.so.2`, `libtensorflow_framework.so.2`, and `_pywrap_tensorflow_internal.so` symlinks in the `sys._MEIPASS` cause the discovery code to search in wrong directories (i.e., it takes the unresolved location of these files as base location for the search), as removing those three symlinks from generated frozen application seems to fix things.

To deal with this, we will need to add a mechanism that will allow the hooks to suppress (specific) symlinks generation during binary dependency analysis; but I need to think whether to introduce a dedicated mechanism for that, or try to make it part of more general file-exclusion mechanism that we were discussing some time back...

bwoodsend (approved):

oushu1zhangxiangxuan1:
How do we use these hooks? If it works as defaults?

bwoodsend:
If you have `pyinstaller-hooks-contrib>=2023.12` installed then they will apply automatically.

PR #692, opened 2024-01-18
hooks: add hook for fairscale

rokm:
Add hook for `fairscale` to collect its source .py files for torchscript/JIT.

Add a basic import test that demonstrates the need for that.


bwoodsend:
No new entry in `requirements-test-libraries.txt`?

rokm:
Nope - it's another `torch` based package, and we're not bogging up our CI with that.

bwoodsend (approved):

PR #693, opened 2024-01-22
Improve osgeo hook

eggertm:
Had issues trying to use GDAL (using the unofficial windows binaries), when loading a chart with EPSG coordinates.  This required the data/proj folder to be bundled as well.

### DIFF
--- a/news/693.update.rst
+++ b/news/693.update.rst
@@ -1,0 +1,1 @@
+    Fix hook for ``osgeo``, to include proj data files.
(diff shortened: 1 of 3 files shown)


eggertm:
The CI test appears to fail on running cython, which I don't think is related to any of my rather miniscule changes.

rokm:
By "unofficial windows binaries", you mean the builds at https://www.lfd.uci.edu/~gohlke/pythonlibs/#gdal ?

eggertm:
Yes, correct.

rokm (approved):

PR #703, opened 2024-02-15
hooks: add hook for falcon

### LINKED ISSUE #702: Falcon hook

`falcon`

Yes, but it needs several hidden imports.

**Additional context**

See also: https://github.com/falconry/falcon/discussions/2068

PR incoming...

davetapley:
closes #702

### DIFF
--- a/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-falcon.py
+++ b/src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-falcon.py
@@ -1,0 +1,27 @@
+# ------------------------------------------------------------------
+# Copyright (c) 2024 PyInstaller Development Team.
+#
+# This file is distributed under the terms of the GNU General Public
+# License (version 2.0 or later).
+#
+# The full license is available in LICENSE.GPL.txt, distributed with
+# this software.
+#
+# SPDX-License-Identifier: GPL-2.0-or-later
+# ------------------------------------------------------------------
+
+hiddenimports = [
+    'cgi',
+    'falcon.app_helpers',
+    'falcon.forwarded',
+    'falcon.media',
+    'falcon.request_helpers',
+    'falcon.responders',
+    'falcon.response_helpers',
+    'falcon.routing',
+    'falcon.vendor.mimeparse',
+    'falcon.vendor',
+    'uuid',
+    'xml.etree.ElementTree',
+    'xml.etree'
+]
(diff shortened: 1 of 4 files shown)


bwoodsend:
Hmm, it looks like having both the original `.py` files and the cythonized `.so` files available is somehow causing neither to be found. Running `rm env/lib/python3.12/site-packages/falcon/**.py` makes the tests pass. There's presumably a PyInstaller bug in here although, depending on where it comes from, it may not actually be fixable. Grrr, there's a special place in hell for packages which optionally squirt themselves through cython instead of just cythonizing the bottlenecks...

src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-falcon.py
```diff
@@ -0,0 +1,27 @@
...
+# License (version 2.0 or later).
+#
+# The full license is available in LICENSE.GPL.txt, distributed with
+# this software.
+#
+# SPDX-License-Identifier: GPL-2.0-or-later
+# ------------------------------------------------------------------
+
+hiddenimports = [
+    'cgi'
```
rokm, line 14: There's a comma missing here at the end, which causes the the first hidden import to become `cgifalcon.app_helpers`, and hence we miss both `cgi` and `falcon.app_helpers`.

rokm:
> Running `rm env/lib/python3.12/site-packages/falcon/**.py` makes the tests pass.

Only because `falcon.__init__` is not cythonized, and so the package becomes a namespace package, whose import is a no-op.

But on the other hand, since `falcon.__init__` is not cythonized, we should automatically pick up the `falcon.app_helpers` that it imports...

rokm:
> But on the other hand, since `falcon.__init__` is not cythonized, we should automatically pick up the `falcon.app_helpers` that it imports...

Actually, scratch that. The `falcon.app_helpers` is imported in `falcon.app`, which is cythonized.

src/_pyinstaller_hooks_contrib/hooks/stdhooks/hook-falcon.py
davetapley: Well spotted! I know exactly what happened: I used the VSCode 'Sort' function and TIL it's not smart enough to fix the commas 🙄

davetapley:
@bwoodsend @rokm the cython discussion is a bit over my head, but I fixed the `,`, let me know if there's anything else I can do 😁

rokm (approved):

bwoodsend:
Thanks @davetapley!

PR #713, opened 2024-03-08
hooks: torch: add support for MKL-enabled torch builds

rokm:
Add support for MKL-enabled `torch` builds on Windows (e.g., the nightly `2.3.0.dev20240308+cpu` build). The `torch` hook now attempts to discover and collect DLLs from MKL and its dependencies (`mkl`, `tbb`, `intel-openmp`).

This should prevent frozen program from silently crashing due to missing MKL DLLs, which slip past PyInstaller's binary dependency analysis due to being dynamically loaded at run time.

Closes pyinstaller/pyinstaller#8348.

### DIFF
--- a/news/712.update.rst
+++ b/news/712.update.rst
@@ -1,0 +1,4 @@
+Update ``torch`` hook to add support for MKL-enabled ``torch`` builds
+on Windows (e.g., the nightly ``2.3.0.dev20240308+cpu`` build). The hook
+now attempts to discover and collect DLLs from MKL and its dependencies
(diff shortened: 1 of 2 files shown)


bwoodsend (approved):
Who doesn't want an extra half a gigabyte added to their application's footprint... :roll_eyes:

rokm:
> Who doesn't want an extra half a gigabyte added to their application's footprint... 🙄

A small price to pay for CPU-powered (or Intel GPU-powered, presumably) chatbots being *faster* at spewing inaccuracies...

PR #742, opened 2024-05-09
hooks: add hook for schwifty

rokm:
Add hook for `schwfity` that collects the data files (the JSON files in `bank_registry`) and package's metadata.

### DIFF
--- a/news/742.new.rst
+++ b/news/742.new.rst
@@ -1,0 +1,2 @@
+Add hook for ``schwifty``. Requires ``schwifty >= 2024.5.1`` due to
+issues with data search path in earlier versions.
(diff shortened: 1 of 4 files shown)


mdomke:
It seems as if the tests fail, because `schwifty` does not include `typing-extensions` as dependency. This must have slipped through my own tests, because the library probably installed by some test-dependency. I will create a new release that fixe this problem.

rokm:
> It seems as if the tests fail, because `schwifty` does not include `typing-extensions` as dependency. This must have slipped through my own tests, because the library probably installed by some test-dependency. I will create a new release that fixe this problem.

Indeed - I've opened https://github.com/mdomke/schwifty/issues/203.

mdomke:
Please update to `2024.05.2` where the issue has been fixed and `typing-extensions` is explicitly required again for Python < 3.11.

bwoodsend (approved):

PR #746, opened 2024-05-18
Improve support for cfgrib

### LINKED ISSUE #744: A problem converting a grib2 file to a netCDF file using cfgrib and xarray
My test code - 
```
import cfgrib
import xarray as xr

# Path to the GRIB2 file
grib_file = 'GFS_Wind_000.grib2'

# Path for the output netCDF file
netcdf_file = 'output.nc'

# Open the GRIB2 file using cfgrib
ds = cfgrib.open_dataset(grib_file)

# Convert the GRIB2 dataset to xarray dataset
xds = xr.Dataset.from_dataframe(ds.to_dataframe())

# Save the xarray dataset to netCDF format
xds.to_netcdf(netcdf_file)
```

this works in python from the terminal window, but when I compile the code using the following pyinstaller command line -
```
pyinstaller --onefile --copy-metadata numpy --collect-data gribapi TryGrib.py
```
and then run the exe I get the following error message - 
```
  File "TryGrib.py", line 11, in <module>
    ds = cfgrib.open_dataset(grib_file)
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "cfgrib\xarray_store.py", line 39, in open_dataset
  File "xarray\backends\api.py", line 557, in open_dataset
  File "xarray\backends\plugins.py", line 205, in get_backend
ValueError: unrecognized engine cfgrib must be one of: ['netcdf4', 'scipy', 'store']
```

any help will be greatly appreciated

rokm:
Extend the `xarray` hook to collect additional backend plugins that are registered via the `xarray.backends` entry-point (e.g., `cfgrib`).

Add hook for `gribapi` package from `eccodes` dist, in order to collect bundled headers and ensure that the eccodes shared library is collected from the build environment.

Manually tested on Windows with Anaconda with program that tries to read a grib file. 

The `test_eccodes_gribapi` might pass on linux, but in reality, the collected shared library cannot yet be discovered (and so the system copy is used). For this to properly work, we will need to add an override for POSIX implementation of `ctypes.util.find_library` in PyInstaller's bootstrap/loader, which would explicitly search for matching filenames in `sys._MEIPASS`.

Closes #744.

### DIFF
--- a/.github/workflows/pr-test.yml
+++ b/.github/workflows/pr-test.yml
@@ -100,6 +100,8 @@ jobs:
           # These are dependencies of gmsh
           sudo apt-get install -y libglu1 libgl1 libxrender1 libxcursor1 libxft2 \
             libxinerama1 libgomp1
+          # This one is required by eccodes
+          sudo apt-get install -y libeccodes0
 
       - name: Install brew dependencies
         if: startsWith(matrix.os, 'macos')
@@ -112,6 +114,8 @@ jobs:
           brew install libdiscid
           # Install lsl library for pylsl
           brew install labstreaminglayer/tap/lsl
+          # This one is required by eccodes
+          brew install eccodes
 
       - name: Install dependencies
         shell: bash
--- a/.github/workflows/pr-test.yml
+++ b/.github/workflows/pr-test.yml
@@ -43,9 +43,11 @@ jobs:
         run: |
           git fetch origin ${{ github.base_ref }}
           set +e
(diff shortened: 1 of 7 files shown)


rokm:
Hmmm, looks like `macos-latest` now resolves to Sonoma on arm64, and our dependency diff is broken because that ships Apple diff instead of GNU one...

bwoodsend (approved):

PR #749, opened 2024-05-23
Update LICENSE.APL.txt

kcehvoin1:
Add full text of Apache 2.0 License - copied from https://www.apache.org/licenses/LICENSE-2.0.


LICENSE.APL.txt
```diff
@@ -22,6 +22,193 @@ though please note that this license only applies to files which state so in the
...
+      To apply the Apache License to your work, attach the following
+      boilerplate notice, with the fields enclosed by brackets "[]"
+      replaced with your own identifying information. (Don't include
+      the brackets!)  The text should be enclosed in the appropriate
+      comment syntax for the file format. We also recommend that a
+      file or class name and description of purpose be included on the
+      same "printed page" as the copyright notice for easier
+      identification within third-party archives.
 
    Copyright 2020-2021 PyInstaller Development Team
```
rokm, line 213: Technically, this part down here was previously the actual copyright notice, based on the boilerplate from the Appendix with []-enclosed fields filled in. Now, it kind of looks like it was subsumed back into the boilerplate notice, but with fields filled in, which looks a bit weird.

I suppose if we want to retain the actual notice part here, it could be moved to the top, above the full license text, while the license contains the unmodified template for the sake of consistency.

kcehvoin1:
Sorry for the messy commit history, new to GitHub Codespace's UI.

rokm:
> The file should probably be renamed too. APL (what it's currently called) is _Adaptable Public License_. Apache's SPDX identifier is Apache-2.0.

Good point. 

There also seems to be a bit of chaos in the hooks themselves - several standard hooks claim to be under Apache-2.0 (which seems like an oversight to me, although it *could* be intentional).

The run-time hooks seem to be referring to `COPYING.txt` for license instead of the `LICENSE.txt` or `LICENSE.APL.txt`. That's probably a left-over from when they were moved from the main repository.

But I suppose we could merge the licenses into a single license file, like we have in the main repository? Then we do not have to worry about the naming of the individual files...

bwoodsend:
> But I suppose we could merge the licenses into a single license file, like we have in the main repository? Then we do not have to worry about the naming of the individual files...

I guess someone might feel legally compelled to bundle the runtime hooks' license and exclude the GPL license if their interpretation of _Source form_ includes bytecode.

rokm (approved):

PR #760, opened 2024-07-11
hooks: hydra: add work-around for python < 3.10 and PyInstaller >= 5.8

rokm:
Under python < 3.10, `hydra`'s plugin manager continues to use deprecated PEP-302 functionality that was removed from PyInstaller's `PyiFrozenImporter` in PyInstaller 5.8 (pyinstaller/pyinstaller#7344.). At run-time, this results in `AttributeError: 'PyiFrozenImporter' object has no attribute 'find_module'`.

As a work-around, check python version and PyInstaller version, and if using python < 3.10 and PyInstaller >= 5.8, set module collection mode for `hydra._internal.core_plugins` and `hydra_plugins` packages to `py` to collect their modules as source .py files only. This way, they end up handled by python's built-in finder/importer, instead of PyInstaller's `PyiFrozenImporter`.

### DIFF
--- a/news/760.update.rst
+++ b/news/760.update.rst
@@ -1,0 +1,8 @@
+Update ``hydra`` hook to include work-around for ``hydra``'s plugin
+manager, which under python < 3.10 (still) uses deprecated PEP-302
+that was removed from PyInstaller's ``PyiFrozenImporter`` in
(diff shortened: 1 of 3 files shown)


rokm:
Oneshot test: https://github.com/rokm/pyinstaller-hooks-contrib/actions/runs/9898673929

bwoodsend (approved):

PR #763, opened 2024-07-15
Fix release tag/version management

bwoodsend:
* Create a release tag after generating the changelog so that `https://github.com/pyinstaller/pyinstaller-hooks-contrib/archive/refs/heads/$version.zip` and `https://github.com/pyinstaller/pyinstaller-hooks-contrib/blob/$version/CHANGELOG.rst` contain the right changelog entries
* Similarly, create sdists after generating the changelog so that they also contain the current release's changelog entries
* Bump the project version immediately before release rather than after (for convenience of implication rather than any functional reason)
* Rename GitHub releases from `XXXX.X` to `vXXXX.X` to match the tags. This in turn fixes GitHub generating additional tags without the `v` suffix

Verification [run](https://github.com/bwoodsend/pyinstaller-hooks-contrib/actions/runs/9946733033) and [mock release](https://github.com/bwoodsend/pyinstaller-hooks-contrib/releases/tag/v2024.8)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -25,32 +25,32 @@ jobs:
       - name: Install Dependencies
         run: python -m pip install --upgrade -r requirements-release.txt
 
+      - name: Bump version
+        run: python setup.py bump --build
+
+      - name: Build Changelog
+        run: python -m towncrier --yes
+
       - name: Build & Validate Bundles
         run: |
           python setup.py sdist bdist_wheel
           python -m twine check dist/*
-
-      - name: Build Changelog
-        run: python -m towncrier --yes
 
       - name: Upload to PyPI
         env:
           TWINE_PASSWORD: ${{ secrets.TWINETOKEN }}
         run: python -m twine upload dist/* --disable-progress-bar -u __token__ --non-interactive
 
-      - name: Create Git Tag
-        run: git tag v$(python setup.py --version)
-
-      - name: Bump version, and commit changes back into the repository
+      - name: Commit changes and publish to GitHub
         run: |
-          VERSION=$(python setup.py --version)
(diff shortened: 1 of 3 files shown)


bwoodsend:
> Rename GitHub releases from XXXX.X to vXXXX.X to match the tags. This in turn fixes GitHub generating additional tags without the v suffix

Or we could keep the mismatch and just tell `gh release` to use the `v` prefixed tag?

rokm (approved):

rokm:
> > Rename GitHub releases from XXXX.X to vXXXX.X to match the tags. This in turn fixes GitHub generating additional tags without the v suffix
> 
> Or we could keep the mismatch and just tell `gh release` to use the `v` prefixed tag?

Either approach is fine with me.

rokm:
Looks like `setup.cfg` is overdue for some cleanup as well: https://github.com/bwoodsend/pyinstaller-hooks-contrib/actions/runs/9946733033/job/27477914681#step:6:15

bwoodsend:
Hmm, it's scary how much my brain is trained to ignore everything that setuptools prints following years of ignoring all the unactionable `setup.py xxx is deprecated` warnings.

bwoodsend:
There are [these](https://github.com/bwoodsend/pyinstaller-hooks-contrib/actions/runs/9946733033/job/27477914681#step:8:1115) floating around too... :disappointed:

bwoodsend:
Can we move the tests out of `src`? I never liked that out test suite end up inthe PyPI wheels – especially given that I'm always criticising other packages for doing the same.

rokm:
> Can we move the tests out of `src`?

Yeah, I was thinking the same thing...

PR #764, opened 2024-07-16
Assorted setup/packaging updates

bwoodsend:
Following on from https://github.com/pyinstaller/pyinstaller-hooks-contrib/pull/763#issuecomment-2229520651 plus a couple of other things that annoyed me when I saw them.


rokm:
1. Is there any benefit of keeping the `src` directory and not moving `_pyinstaller_hooks_contrib` into top-level directory (to keep layout consistent with the main repo)?

2. The `hooks` directory restructuring will break `hook-nvidia.*` that were using `_pyinstaller_hooks_contrib.hooks.util` (e.g., [here](https://github.com/bwoodsend/pyinstaller-hooks-contrib/blob/packaging-updates/src/_pyinstaller_hooks_contrib/stdhooks/hook-nvidia.cublas.py)), although in retrospect, it was my fault for creating `_pyinstaller_hooks_contrib.hooks.util`  instead of `_pyinstaller_hooks_contrib.util` in the first place.

3. An alternative restructuring of the `hooks` directory would be to keep the `hooks` directory, and move all standard hooks from `stdhooks` into it, while keeping the `pre_find_module_path`, `pre_safe_import_module`, and `rthooks` as subdirectories under `hooks`. This way, we would have "standard" hooks directory layout and not have to [return extra path](https://github.com/bwoodsend/pyinstaller-hooks-contrib/blob/35c23af8f0055110aae52e24d53b897068d57f66/src/_pyinstaller_hooks_contrib/__init__.py#L22-L23). (The `_pyinstaller_hooks_contrib.hooks.util` should probably be renamed to `_pyinstaller_hooks_contrib.util` anyway).

bwoodsend:
> Is there any benefit of keeping the src directory and not moving _pyinstaller_hooks_contrib into top-level directory (to keep layout consistent with the main repo)?

It keeps other directories out of `sys.path` (mostly the `tests` directory) for editable installs and gives setuptools a hint not to collect them. Neither of which bother me so I'd be in favour of collapsing that directory away too.

> An alternative restructuring of the hooks directory would be to keep the hooks directory, and move all standard hooks from stdhooks into it

Hmm, I quite like having the standard hooks in their own bucket (even if it does make `get_hook_dirs()` look a bit silly).

rokm (approved):

PR #775, opened 2024-08-06
Add hooks for trame

### LINKED ISSUE #773: Trame suite hooks

[All Trame libraries](https://github.com/Kitware?q=trame-). Examples:

- `trame`
- `trame-client`
- `trame-components`
- `trame-datagrid`
- `trame-server`
- `trame-tauri`
- `trame-vtk`
- `trame-vuetify`
- `trame-xterm`

Yes, by creating a hook for each library with the line `datas = collect_data_files("LIBRARY_NAME", subdir="module")`.

**Additional context**

Most (maybe all) Trame libraries have a data folder called `module` that needs to be collected. I was planning to create a bunch of PRs for each repo, but I thought maybe there would be a good way to get all the libraries in one-shot to future proof against new Trame libraries as they frequently come out.

MattTheCuber:
Closes #773 

I manually tested **all** libraries (except `vtklocal` because it is unstable and currently broken) that I added, and created a couple library tests. I wanted to add tests for every package (and did). However, trame is a client-server model framework that starts a server when you run it. I am not sure how to or what to even mock to disable running a server in tests. For now, I removed those tests. If you are fine with starting and stopping servers I can add back the tests. Maybe @jourdain (creator of Trame) has a better idea for writing tests for this?

Here is an example test (which does pass on my machine):
```python
import asyncio

from trame.app import get_server
from trame.ui.html import DivLayout
from trame_xterm.widgets import xterm


async def stop(*args, **kwargs):
    await server.stop()


server = get_server()
with DivLayout(server):
    xterm.XTerm()
server.controller.on_server_ready.add(
    lambda *args, **kwargs: asyncio.ensure_future(stop(*args, **kwargs))
)
server.start(port=5102, open_browser=False)
```

### DIFF
--- a/tests/test_trame.py
+++ b/tests/test_trame.py
@@ -36,7 +36,7 @@ def test_trame_client(pyi_builder):
         server.controller.on_server_ready.add(
             lambda *args, **kwargs: asyncio.ensure_future(stop(*args, **kwargs))
         )
-        server.start(port=5100, open_browser=False)
+        server.start(port=0, open_browser=False)
     """)
 
 
@@ -58,7 +58,7 @@ def test_trame_vuetify(pyi_builder):
         server.controller.on_server_ready.add(
             lambda *args, **kwargs: asyncio.ensure_future(stop(*args, **kwargs))
         )
-        server.start(port=5101, open_browser=False)
+        server.start(port=0, open_browser=False)
     """)
 
 
@@ -67,8 +67,9 @@ def test_trame_vuetify(pyi_builder):
 def test_trame_vtk(pyi_builder):
     pyi_builder.test_source("""
         import os
-        import trame_vtk
+
         import pyvista as pv
+        import trame_vtk
 
         plotter = pv.Plotter()
         plotter.export_html("test.html")  # Uses trame_vtk
--- a/tests/test_trame.py
+++ b/tests/test_trame.py
@@ -1,0 +1,631 @@
+# ------------------------------------------------------------------
+# Copyright (c) 2020 PyInstaller Development Team.
+#
+# This file is distributed under the terms of the GNU General Public
+# License (version 2.0 or later).
+#
+# The full license is available in LICENSE, distributed with
+# this software.
+#
+# SPDX-License-Identifier: GPL-2.0-or-later
+# ------------------------------------------------------------------
+
+from PyInstaller.utils.tests import importorskip
+
+
+@importorskip("trame")
+def test_trame(pyi_builder):
+    pyi_builder.test_source("""
+        import trame
+    """)
+
+
+@importorskip('trame_client')
+def test_trame_client(pyi_builder):
+    pyi_builder.test_source("""
+        import asyncio
+
+        from trame.app import get_server
+
+
+        async def stop(*args, **kwargs):
+            await server.stop()
+
+
+        server = get_server()
+        server.controller.on_server_ready.add(
+            lambda *args, **kwargs: asyncio.ensure_future(stop(*args, **kwargs))
+        )
+        server.start(port=5100, open_browser=False)
+    """)
+
+
+@importorskip('trame_vuetify')
+def test_trame_vuetify(pyi_builder):
+    pyi_builder.test_source("""
+        import asyncio
+
+        from trame.app import get_server
+        from trame.ui.vuetify3 import VAppLayout
+
+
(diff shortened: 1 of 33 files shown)


MattTheCuber:
Do I need to add the hypen versions too (`trame-components`)?

rokm:
> Do I need to add the hypen versions too (e.g. `hook-trame-components.py`)?

No, PyInstaller's hooks are tied to importable package/module names, not dist names.

rokm:
Looks like `test_trame_vtk` also needs an `@importorskip('pyvista')`.

rokm:
Does the lack of data files from those packages manifest itself as an error when a package is imported, or only when trying to start the server? If the former, we could go with just basic import tests. Otherwise, I don't see a problem with starting and stopping a server, either; but the port will probably need to be auto-selected (in case two such tests are run in parallel, or in case that hard-coded port is taken for whatever reason).

rokm:
If you are going to add multiple `trame`-related tests, maybe put them into separate `test_trame.py` file instead of `test_libraries.py`.

MattTheCuber:
> Does the lack of data files from those packages manifest itself as an error when a package is imported, or only when trying to start the server?

They are not loaded on import. The only time I have been able to make them load is when the server is started.

> the port will probably need to be auto-selected

How would I do this? I was planning to just hard code a different port for each test.

bwoodsend:
>>  the port will probably need to be auto-selected

> How would I do this? I was planning to just hard code a different port for each test.

Does setting the port to 0 do the job?

MattTheCuber:
> > > the port will probably need to be auto-selected
> 
> > How would I do this? I was planning to just hard code a different port for each test.
> 
> Does setting the port to 0 do the job?

Good question. I don't know, I will try that.

jourdain:
Yes using port 0 will work. And you do indeed to start the server.

MattTheCuber:
All of these tests pass locally on my machine. Not sure why they are failing in the CI?

rokm:
> All of these tests pass locally on my machine. Not sure why they are failing in the CI?

The failing test needs import check for `vtkmodules`.

tests/test_trame.py
```diff
@@ -412,6 +412,7 @@ async def stop(*args, **kwargs):
     """)
 
 
+@importorskip("vtk")
```
rokm, line 415: `vtk` is dist name, importable package is `vtkmodules` (as in the error message of the failed test).

MattTheCuber:
Btw, this is still a draft because there are so many other libraries that need hooks that I haven't actually tested my program works 😄

MattTheCuber:
Can you take a look at the `test_trame_vtk_tools` test? I can't seem to make it pass in the CI or locally.

rokm:
Yeah, looks like import check for `nest_asyncio` is missing (and `nest-asyncio` needs to be added to requirements).

MattTheCuber:
How did you find that?

rokm:
> How did you find that?

Check the traceback from the failed test's executable run:
https://github.com/pyinstaller/pyinstaller-hooks-contrib/actions/runs/10271758222/job/28422471043#step:10:973

tests/test_trame.py
```diff
@@ -0,0 +1,660 @@
+# ------------------------------------------------------------------
+# Copyright (c) 2020 PyInstaller Development Team.
```
bwoodsend, line 2: :slightly_smiling_face:

tests/test_trame.py
```diff
@@ -0,0 +1,660 @@
...
+def test_trame_vtk_tools(pyi_builder):
+    pyi_builder.test_source("""
+        import os
+
+        import pyvista as pv
+        import trame_vtk
+
+        plotter = pv.Plotter()
+        plotter.export_html("test.html")  # Uses trame_vtk
+        os.remove("test.html")
```
bwoodsend, line 103: Can we use pytest's `tmp_path` fixture here and put the exported file under there?

bwoodsend (commented):
Looks good...

bwoodsend (approved):

rokm (approved):

bwoodsend:
Let me know if/when you've finished your pull request binge and I'll hit the release button.

MattTheCuber:
I believe I'm all set for now, our application seems to be running smoothly now. Thanks!

PR #784, opened 2024-08-26
Add hook for slixmpp library

15498th:
Fix runtime import error for [slixmpp](https://codeberg.org/poezio/slixmpp) Jabber library.

Modules in "features" are dynamically imported in the same way plugins are, however, unlike plugins imports, they are not detected automatically (probably due to the fact that import happens at runtime, when the client class is instantiated), leading to the following error:

```
Unable to load plugin: feature_starttls
  File "slixmpp\plugins\base.py", line 78, in load_plugin
ModuleNotFoundError: No module named 'slixmpp.plugins.feature_starttls'

During handling of the above exception, another exception occurred:

  File "slixmpp\plugins\base.py", line 82, in load_plugin
ModuleNotFoundError: No module named 'slixmpp.features'
Unable to load plugin: feature_starttls
  File "slixmpp\plugins\base.py", line 78, in load_plugin
ModuleNotFoundError: No module named 'slixmpp.plugins.feature_starttls'

During handling of the above exception, another exception occurred:

  File "slixmpp\plugins\base.py", line 82, in load_plugin
ModuleNotFoundError: No module named 'slixmpp.features'
  File "test_source.py", line 10, in <module>
  File "test_source.py", line 8, in __init__
  File "slixmpp\clientxmpp.py", line 122, in __init__
  File "slixmpp\basexmpp.py", line 269, in register_plugin
  File "slixmpp\plugins\base.py", line 153, in enable
slixmpp.plugins.base.PluginNotFound: feature_starttls
```

### DIFF
--- a/tests/test_libraries.py
+++ b/tests/test_libraries.py
@@ -2150,3 +2150,17 @@ def test_itk(pyi_builder):
     pyi_builder.test_source("""
         import itk
     """)
+
+
+@importorskip('slixmpp')
+def test_slixmpp(pyi_builder):
+    pyi_builder.test_source("""
+        import slixmpp
+
+
+        class JabberClient(slixmpp.ClientXMPP):
+            def __init__(self):
+                super().__init__('username', 'password')
+
+        _ = JabberClient()
+    """)
--- a/tests/test_libraries.py
+++ b/tests/test_libraries.py
@@ -2158,9 +2158,5 @@ def test_slixmpp(pyi_builder):
         import slixmpp
 
 
-        class JabberClient(slixmpp.ClientXMPP):
-            def __init__(self):
-                super().__init__('username', 'password')
-
-        _ = JabberClient()
+        slixmpp.ClientXMPP('username', 'password')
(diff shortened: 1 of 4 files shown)


tests/test_libraries.py
```diff
@@ -2150,3 +2150,17 @@ def test_itk(pyi_builder):
...
     """)
+
+
+@importorskip('slixmpp')
+def test_slixmpp(pyi_builder):
+    pyi_builder.test_source("""
+        import slixmpp
+
+
+        class JabberClient(slixmpp.ClientXMPP):
```
rokm, line 2161: Why not instantiate `slixmpp.ClientXMPP('username', 'password')` directly?

15498th: This is how it gets used in real code, base class doesn't do much without some methods overwritten.
For this specific test it does make no difference, I suppose.

rokm (approved):

rokm:
Thank you!

PR #787, opened 2024-08-31
Add hook for capstone

arossert:
Add hook for `capstone` to collect needed libs


_pyinstaller_hooks_contrib/stdhooks/hook-capstone.py
```diff
@@ -0,0 +1,15 @@
+# ------------------------------------------------------------------
+# Copyright (c) 2021 PyInstaller Development Team.
```
rokm, line 2: 2024

PR #796, opened 2024-09-10
hooks: update pydicom hook for compatibility with pydicom v3.0.0

### LINKED ISSUE #795: Hidden imports not found with pydicom 3
**Describe the bug**
* Which hook/library isn't working? `pydicom`
* Does the error get raised while building or when running?
Warning when building, error when running.

**To Reproduce**

A minimal example file:
```python
from pydicom import dcmread
dcmread("")
```

PyInstaller command:
```
pyinstaller test.py
```

Error:
```
(.venv) C:\Code\project>dist\test\test.exe
  File "C:\ws\project\test.py", line 1, in <module>
    from pydicom import dcmread
  File "PyInstaller\loader\pyimod02_importers.py", line 378, in exec_module
  File "pydicom\__init__.py", line 31, in <module>
  File "PyInstaller\loader\pyimod02_importers.py", line 378, in exec_module
  File "pydicom\dataelem.py", line 17, in <module>
  File "PyInstaller\loader\pyimod02_importers.py", line 378, in exec_module
  File "pydicom\config.py", line 367, in <module>
  File "PyInstaller\loader\pyimod02_importers.py", line 378, in exec_module
  File "pydicom\pixel_data_handlers\__init__.py", line 6, in <module>
  File "PyInstaller\loader\pyimod02_importers.py", line 378, in exec_module
  File "pydicom\pixels\__init__.py", line 3, in <module>
  File "PyInstaller\loader\pyimod02_importers.py", line 378, in exec_module
  File "pydicom\pixels\decoders\__init__.py", line 3, in <module>
  File "PyInstaller\loader\pyimod02_importers.py", line 378, in exec_module
  File "pydicom\pixels\decoders\base.py", line 1805, in <module>
  File "pydicom\pixels\common.py", line 130, in add_plugins
  File "pydicom\pixels\common.py", line 83, in add_plugin
  File "importlib\__init__.py", line 90, in import_module
ModuleNotFoundError: No module named 'pydicom.pixels.decoders.gdcm'
[PYI-32016:ERROR] Failed to execute script 'test' due to unhandled exception!
```

**Expected behavior**
No `ModuleNotFoundError`

 - OS: Windows
 - Python Version: Python 3.12
 - Version of `pyinstaller-hooks-contrib`: 2024.8
 - Version of PyInstaller: 6.10.0

rokm:
Update the hidden imports for renamed and extended set of plugin modules, and collect the data files that are now required at run-time due to `pydicom.examples` being imported during `pydicom` module initialization.

Closes #795.

### DIFF
--- a/_pyinstaller_hooks_contrib/stdhooks/hook-pydicom.py
+++ b/_pyinstaller_hooks_contrib/stdhooks/hook-pydicom.py
@@ -10,8 +10,70 @@
 # SPDX-License-Identifier: GPL-2.0-or-later
 # ------------------------------------------------------------------
 
-hiddenimports = [
-    "pydicom.encoders.gdcm",
-    "pydicom.encoders.pylibjpeg",
-    "pydicom.encoders.native",
-]
+from PyInstaller.utils.hooks import is_module_satisfies, collect_data_files
+
+hiddenimports = []
+datas = []
+
+# In pydicom 3.0.0, the `pydicom.encoders` plugins were renamed to `pydicom.pixels.encoders`, and
+# `pydicom.pixels.decoders` were also added. We need to collect them all, because they are loaded during
+# `pydicom` module initialization. We intentionally avoid using `collect_submodules` here, because that causes
+# import of `pydicom` with logging framework initialized, which results in error tracebacks being logged for all plugins
+# with missing libraries (see https://github.com/pydicom/pydicom/issues/2128).
+if is_module_satisfies('pydicom >= 3.0.0'):
+    hiddenimports += [
+        "pydicom.pixels.decoders.gdcm",
+        "pydicom.pixels.decoders.pylibjpeg",
+        "pydicom.pixels.decoders.pillow",
+        "pydicom.pixels.decoders.pyjpegls",
+        "pydicom.pixels.decoders.rle",
+        "pydicom.pixels.encoders.gdcm",
+        "pydicom.pixels.encoders.pylibjpeg",
+        "pydicom.pixels.encoders.native",
+        "pydicom.pixels.encoders.pyjpegls",
+    ]
+
+    # With pydicom 3.0.0, initialization of `pydicom` (unnecessarily) imports `pydicom.examples`, which attempts to set
+    # up several test datasets: https://github.com/pydicom/pydicom/blob/v3.0.0/src/pydicom/examples/__init__.py#L10-L24
+    additional_data_patterns = [
+        # `pydicom/data/urls.json` and `pydicom/data/hashes.json` are required for data files that need to be downloaded
+        # at run-time. The lack of former results in run-time error, while the lack of latter results in warnings about
+        # dataset download failure.
+        'urls.json',
+        'hashes.json',
+        # Try to collect these from the package, to avoid having to download them at run-time. The `pydicom` searches
+        # for these files via recursive glob (`pathlib.Path.rglob`) to allow them being in sub-directories (for example,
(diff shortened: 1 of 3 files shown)


bersbersbers:
Works great for me, thanks!

bwoodsend:
I think we're going to have to make a fuss upstream about those examples. It looks to me like they wouldn't exist at build time in a fresh (e.g. CI/CD) environment so the resultant application would have to redownload everything every time in onefile mode, blow up the macOS sandbox and run afoul of permission errors if installed in a read only location. I'll have a proper look (and possible moan) tomorrow when I'm more awake.

rokm:
> I think we're going to have to make a fuss upstream about those examples. It looks to me like they wouldn't exist at build time in a fresh (e.g. CI/CD) environment so the resultant application would have to redownload everything every time in onefile mode, blow up the macOS sandbox and run afoul of permission errors if installed in a read only location. I'll have a proper look (and possible moan) tomorrow when I'm more awake.

On linux, the downloaded files seem to go into `~/.pydicom/data`, so this does not violate the immutability of the frozen application bundle. But I agree that it is a bit of a WTF, especially since there is no reason for [this import](https://github.com/pydicom/pydicom/blob/b2073cccbd92d537dfb307c4d351a267e602069e/src/pydicom/__init__.py#L33) to exist.

bwoodsend:
I guess that's not quite so bad then. I raised https://github.com/pydicom/pydicom/issues/2131 anyway. Is it worth putting an `import pydicom` into the hook to avoid the *examples not packaged because the application was built on a clean CI machine* scenario?

rokm:
> Is it worth putting an `import pydicom` into the hook to avoid the _examples not packaged because the application was built on a clean CI machine_ scenario?

I don't think so - some of those files are part bundled with the package (and we collect those), while some of them are downloaded on first import (and we do not collect those, because they are not part of the package).

bwoodsend:
I think we can get away without `test_files`?

bwoodsend:
Ugh forget it. If I `rm -rf` my `site-packages/pydicom/data/test_files` then the next `import pydicom` doesn't fill `~/.pydicom` with anything new. But reading the code seems to imply it should download them as you say.

bwoodsend (approved):

bwoodsend:
Hopefully that redundant `import pydicom.examples` will disappear.

rokm:
Hmm, you might be right. Because they are not in the `urls.json`, it doesn't actually try to download them.

So the dictionary ends up like this (for frozen application with `test_files` patterns removed):
```
{'ct': None,
 'dicomdir': None,
 'jpeg2k': '/home/rok/.pydicom/data/US1_J2KR.dcm',
 'mr': None,
 'no_meta': None,
 'overlay': '/home/rok/.pydicom/data/MR-SIEMENS-DICOM-WithOverlays.dcm',
 'palette_color': '/home/rok/.pydicom/data/OBXXXX1A.dcm',
 'rgb_color': '/home/rok/.pydicom/data/US1_UNCR.dcm',
 'rt_dose': None,
 'rt_plan': None,
 'rt_ss': None,
 'waveform': None,
 'ybr_color': '/home/rok/.pydicom/data/color3d_jpeg_baseline.dcm'}
```

bwoodsend:
Meh, it's up to you. If https://github.com/pydicom/pydicom/issues/2131 goes where I think it'll go then there's a good chance that we won't need any of them.

rokm:
Let's try getting away without collecting them, then. I've pushed a separate commit so that we have this documented in (the unlikely) event that we end up having to collect them after all.

PR #798, opened 2024-09-12
Add hook for saml2

kfrydel:


_pyinstaller_hooks_contrib/stdhooks/hook-saml2.py
```diff
@@ -0,0 +1,27 @@
...
+# ------------------------------------------------------------------
+
+# Hook for https://github.com/IdentityPython/pysaml2
+from PyInstaller.utils.hooks import collect_data_files, copy_metadata, collect_submodules
+
+datas = copy_metadata("pysaml2")
+
+# The library contains a bunch of XSD schemas that are loaded by the code:
+# https://github.com/IdentityPython/pysaml2/blob/7cb4f09dce87a7e8098b9c7552ebab8bc77bc896
+# /src/saml2/xml/schema/__init__.py#L23
```
rokm, line 20: Can we keep these URLs in single line? Even if it exceeds the nominal line length limit of 120 that we use, I think linter has enough decency to not make a fuss about it when it is a single "word".

kfrydel: Of course. I added a commit with a fix. I was not sure if you prefer new commits or amending.

rokm: >  I was not sure if you prefer new commits or amending.

For PRs like this, we usually end up squashing the commits into a single one during the merge. So either way is fine. (For larger changes that touch multiple places of the codebase, I prefer multiple atomic commits, which I think make the history easier to read).

rokm (approved):

PR #799, opened 2024-09-12
hooks: update gribapi hook for compatibility with eccodes v2.37.0

rokm:
Starting with `eccodes` v2.37.0, upstream provides binary wheels with bundled copy of the `eccodes` shared library. Adjust the hook for `gribapi` (part of `eccodes` dist) to account for that. Also handle the issue with circular imports by importing `eccodes` before `gribapi.bindings` when trying to obtain the path to the shared library.


rokm:
And as usual, I am bitten by making the test a bit more thorough...

bwoodsend (approved):

PR #800, opened 2024-09-12
hooks: xarray: collect xarray.chunkmanagers entry-points

rokm:
Have the `xarray` hook collect `xarray.chunkmanagers` entry-points; this ensures that at least the default (`dask`) chunk manager is detected as present at the run-time.

Closes https://github.com/pyinstaller/pyinstaller/issues/8786.

### DIFF
--- a/.github/workflows/pr-test.yml
+++ b/.github/workflows/pr-test.yml
@@ -70,6 +70,8 @@ jobs:
             echo trame-vuetify >> requirements-test-libraries.txt;
           fi
           if grep -q trame-mesh-streamer requirements-test-libraries.txt; then echo vtk >> requirements-test-libraries.txt; fi
+          # test_xarray_chunk requires dask in addition to xarray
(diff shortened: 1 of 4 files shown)


rokm:
Oneshot test against the latest `xarray` version: https://github.com/rokm/pyinstaller-hooks-contrib/actions/runs/10833342490

rokm:
This time with `dask` installed in addition to `xarray`: https://github.com/rokm/pyinstaller-hooks-contrib/actions/runs/10833638361

Which means I'll need to add another `importorskip` guard to the test, and ensure that `dask` is installed on `xarray` updates.

rokm:
FWIW, there is an `xarray[parallel]` extra, but that in turn pulls in `dask[complete]`, which seems to install more stuff than we need for the test. So I'd rather stick to the usual grep approach.

bwoodsend (approved):

PR #803, opened 2024-09-12
Add a hook for `toga`

davetapley:
https://github.com/beeware/toga

### DIFF
--- a/news/803.new.rst
+++ b/news/803.new.rst
@@ -1,0 +1,1 @@
+Add a hook for ``toga``, which has metadata.
--- a/requirements-test-libraries.txt
+++ b/requirements-test-libraries.txt
@@ -222,6 +222,7 @@ yapf==0.40.2
 xmlschema==2.5.1
 pysaml2==7.5.0; python_version >= '3.9'
 pysaml2==7.3.0; python_version < '3.9'
+toga==0.4.6
 
 # ------------------- Platform (OS) specifics
 
--- a/_pyinstaller_hooks_contrib/stdhooks/hook-toga.py
+++ b/_pyinstaller_hooks_contrib/stdhooks/hook-toga.py
@@ -1,0 +1,30 @@
+# ------------------------------------------------------------------
+# Copyright (c) 2024 PyInstaller Development Team.
+#
+# This file is distributed under the terms of the GNU General Public
+# License (version 2.0 or later).
+#
+# The full license is available in LICENSE, distributed with
+# this software.
+#
+# SPDX-License-Identifier: GPL-2.0-or-later
+# ------------------------------------------------------------------
+
+from PyInstaller.compat import is_darwin, is_linux
+from PyInstaller.utils.hooks import copy_metadata, collect_entry_point
+
+if is_darwin:
+    backend = 'cocoa'
+elif is_linux:
+    backend = 'gtk'
+else:
+    backend = 'winforms'
+
+hiddenimports = [f'toga_{backend}', f'toga_{backend}.factory']
+
+datas = copy_metadata("toga.core") 
+
+for ep in ["toga.backends", "setuptools_scm.parse_scm"]:
+    datas_, hiddenimports_ = collect_entry_point(ep)
+    datas += datas_
+    hiddenimports += hiddenimports_
--- a/tests/test_libraries.py
+++ b/tests/test_libraries.py
@@ -2212,3 +2212,15 @@ def test_saml2(pyi_builder):
         from saml2.attribute_converter import ac_factory
         ac_factory()
     """)
+
+
+@importorskip('toga')
+def test_toga(pyi_builder):
+    pyi_builder.test_source("""
+        import toga
+
+        def build(app):
+            return toga.Box()
+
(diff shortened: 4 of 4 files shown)


requirements-test-libraries.txt
```diff
@@ -222,6 +222,7 @@ yapf==0.40.2
 xmlschema==2.5.1
 pysaml2==7.5.0; python_version >= '3.9'
 pysaml2==7.3.0; python_version < '3.9'
+toga==0.4.6; python_version >= '3.8'
```
rokm, line 225: The python version restrictions should apply to what we currently use on our CI, so `; python_version >= '3.8'` is redundant.

rokm:
Looks like linux runner will require `gir1.2-gtk-3.0` and `libgirepository1.0-dev` to be installed for `PyGObject` (dependency of `toga-gtk`).

news/803.new.rst
```diff
@@ -0,0 +1 @@
+Add a hook for ``toga``, which has metadata.
```
bwoodsend, line 1: All packages *have* metadata. It's really about whether a package feels the need to loop it up at runtime.

rokm: Reckon we should add couple of more examples to complement https://github.com/pyinstaller/pyinstaller-hooks-contrib/blob/d40669c89794e7563c3d0b1f9bace7b353cd58ce/news/README.txt#L38

davetapley:
@rokm @bwoodsend getting to the limit of my knowledge of PyInstaller with this one :|

Couple notes:
- Maintainers of `toga` have their own 'make it a binary' tool [briefcase](https://github.com/beeware/briefcase), so likely aren't going to be _super_ motivated to work on PyInstaller support.

- Whole point of `toga` (and `briefcase`) is to target different OS as well. I know PyInstaller does too, but probably brings a lot more complexity, I started to tackle that here:
  https://github.com/pyinstaller/pyinstaller-hooks-contrib/blob/b2a397fef1f79c5464135917150f8d53d0b4465b/_pyinstaller_hooks_contrib/stdhooks/hook-toga.py#L13-L23

I've got `pyinstaller-hooks-contrib` cloned on a Mac, so targeting that for right now, it uses their `cocoa` engine.

Failing with:

```
 File "toga_cocoa/icons.py", line 21, in __init__
  File "pathlib.py", line 1162, in __init__
  File "pathlib.py", line 373, in __init__
TypeError: argument should be a str or an os.PathLike object where __fspath__ returns a str, not 'NoneType'
```

Probably because we get in to [Objective C bindings](https://github.com/beeware/toga/blob/b0ab9be7d34e85c58d56f4d00b386ed20b28cb52/cocoa/src/toga_cocoa/libs/foundation.py#L18) pretty quick (via [this package](https://github.com/beeware/rubicon-objc/)).

Question: Do you think this will be implementable without a huge amount of effort?

freakboy3742:
> * Maintainers of `toga` have their own 'make it a binary' tool [briefcase](https://github.com/beeware/briefcase), so likely aren't going to be _super_ motivated to work on PyInstaller support.

BeeWare maintainer here! You're correct in guessing that working on PyInstaller support isn't high on our list of priorities. However, that doesn't mean we're *hostile* to PyInstaller. Having more options for installing Toga apps is a good thing, as far as I'm concerned.

To that end - if there's something Toga could do that is preventing it from working with PyInstaller, or there's a change we can make that will make PyInstaller packaging easier, we're not fundamentally opposed to making those changes - especially if that change is based around compliance with existing PEPs. However, it's up to someone else to propose what those changes would be.

rokm:
I've opened a new PR (#804) with additional commits on top of yours, because I do not seem to have permissions to push into this one. Will first try to sort out linux and Windows, though, because they seem more straightforward, and then move on to macOS.

rokm:
And the problem on macOS is that `toga` assumes that the frozen/stand-alone bundled application is always an .app bundle (I assume `briefcase` builds only .app bundles for macOS). 

So it tries to retrieve the .app bundles icon by querying `CFBundleIconFile` key, and passes the result to `pathlib.Path` without checking if is valid (not `None`). So this breaks when you build PyInstaller POSIX executable, but works if you build the .app bundle instead.

I think changing [these lines](https://github.com/beeware/toga/blob/be9bc402a64236abd424f629336f727e32fcfb87/cocoa/src/toga_cocoa/icons.py#L21-L23) from

```python
            bundle_icon = Path(
                NSBundle.mainBundle.objectForInfoDictionaryKey("CFBundleIconFile")
            )
```

to 

```python
            bundle_icon = NSBundle.mainBundle.objectForInfoDictionaryKey("CFBundleIconFile")
            if bundle_icon is None:
                # Not an .app bundle (e.g., POSIX build made with PyInstaller)
                raise FileNotFoundError()
```

should fix it. cc @freakboy3742

freakboy3742:
> And the problem on macOS is that `toga` assumes that the frozen/stand-alone bundled application is always an .app bundle (I assume `briefcase` builds only .app bundles for macOS).

That's correct - and icon handling is one of many reasons for this.

> I think changing 
...
Makes sense. PRs welcome :-)

rokm:
Superseded by #804.

PR #805, opened 2024-09-16
hooks: add hook for setuptools_scm

rokm:
Add hook for `setuptools_scm` to ensure that metadata for `setuptools` dist is collected, in order to avoid the following run-time warning:

```
ERROR: setuptools==0.dev0+unknown is used in combination with setuptools_scm>=8.x

Your build configuration is incomplete and previously worked by accident!
setuptools_scm requires setuptools>=61

Suggested workaround if applicable:
 - migrating from the deprecated setup_requires mechanism to pep517/518
   and using a pyproject.toml to declare build dependencies
   which are reliably pre-installed before running the build tools
```

### DIFF
--- a/_pyinstaller_hooks_contrib/stdhooks/hook-setuptools_scm.py
+++ b/_pyinstaller_hooks_contrib/stdhooks/hook-setuptools_scm.py
@@ -1,0 +1,17 @@
+# ------------------------------------------------------------------
+# Copyright (c) 2024 PyInstaller Development Team.
+#
+# This file is distributed under the terms of the GNU General Public
+# License (version 2.0 or later).
+#
+# The full license is available in LICENSE, distributed with
+# this software.
+#
+# SPDX-License-Identifier: GPL-2.0-or-later
+# ------------------------------------------------------------------
+
+from PyInstaller.utils.hooks import copy_metadata
+
+# Ensure `metadata of setuptools` dist is collected, to avoid run-time warning about unknown/incompatible `setuptools`
+# version.
+datas = copy_metadata('setuptools')
--- a/news/805.new.rst
+++ b/news/805.new.rst
@@ -1,0 +1,3 @@
+Add hook for ``setuptools_scm`` that collects metadata of ``setuptools``
+dist in order to avoid run-time warning about unknown/incompatible
+``setuptools`` version.


rokm:
I've seen the following pattern of a package trying to determine its version a couple of times now:

```python
try:
    from setuptools_scm import get_version
    __version__ = get_version("../..", relative_to=__file__)
except (ModuleNotFoundError, LookupError):
    try:
        from importlib.metadata import version
    except ModuleNotFoundError:
        from importlib_metadata import version

    __version__ = version("<name>")
``` 

In a minimal build environment, `setuptools_scm` should (hopefully?) not be installed, so it is a non-issue. But if it happens to be, the frozen application generates the afore-mentioned run-time warning due to missing `setuptools` metadata. So ensure it is collected.

This is a "last-defense" sort of approach, though. Ideally, we would ensure that `setuptools_scm` is not packaged at all (as it makes no sense in these sort version-determining scenarios).  Perhaps as a future extension, we could have the hook raise an exception when `PYINSTALLER_STRICT_COLLECT_MODE` is enabled, so that we enforce exclusion from all packages that we provide hooks for.

bwoodsend (approved):

bwoodsend:
I'd be surprised of anyone apart from us even knows that `PYINSTALLER_STRICT_COLLECT_MODE` exists so unless we want to use it for this repo to find packages for which we should `excludedimports = ["setuptools_scm"]`, I don't see the point.

Another option might be to create a fake module for `setuptools_scm` which just raises a `ModuleNotFoundError` on startup?

rokm:
> I'd be surprised of anyone apart from us even knows that `PYINSTALLER_STRICT_COLLECT_MODE` exists so unless we want to use it for this repo to find packages for which we should `excludedimports = ["setuptools_scm"]`, I don't see the point.

This was the idea - for us to catch packages where we need  `excludedimports = ["setuptools_scm"]`, without aborting builds for regular users.

> Another option might be to create a fake module for setuptools_scm which just raises a ModuleNotFoundError on startup?

Hmm, indeed. But as far as our luck goes, I wouldn't be surprised if there *is* a legitimate use of `setuptools_scm` in frozen application out in the wild, which would end up broken by this.

bwoodsend:
> 

    I'd be surprised of anyone apart from us even knows that PYINSTALLER_STRICT_COLLECT_MODE exists so unless we want to use it for this repo to find packages for which we should excludedimports = ["setuptools_scm"], I don't see the point.

This was the idea - for us to catch packages where we need excludedimports = ["setuptools_scm"], without aborting builds for regular users.

> > I'd be surprised of anyone apart from us even knows that `PYINSTALLER_STRICT_COLLECT_MODE` exists so unless we want to use it for this repo to find packages for which we should `excludedimports = ["setuptools_scm"]`, I don't see the point.
> 
> This was the idea - for us to catch packages where we need `excludedimports = ["setuptools_scm"]`, without aborting builds for regular users.

Fair enough, sounds sensible.

PR #824, opened 2024-10-22
ci: add python 3.12 and 3.13, and use macos-13 and macos-14

rokm:
Update the CI to cover all supported python versions. Switch from deprecated macos-12 runner to macos-13 (x86_64) and macos-14 (arm64).

### DIFF
--- a/.github/workflows/pr-test.yml
+++ b/.github/workflows/pr-test.yml
@@ -20,9 +20,9 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


rokm:
We skipped the addition of python 3.12 last year, which probably saved us some headaches with packages that were slow to add support for it. But this time, let's try to have full coverage of supported python versions - if it proves to be too much of a hassle, we can disable python 3.13 later.

Similarly, let's enable both x86_64 and arm64 runners for now, and see how that pans out over the next couple of Mondays.

bwoodsend (approved):

PR #826, opened 2024-10-30
Add hook for sv_ttk

aarons20:
This hook ensures that tcl files are collected during the building process.

### DIFF
--- a/news/823.new.rst
+++ b/news/823.new.rst
@@ -1,4 +1,4 @@
 Add hook for ``numbers_parser`` to ensure that package's data file is
 collected.
 
-Add hook for ``sv_ttk`` to ensure that tcl files are collected.
+Add hook for ``sv_ttk`` to ensure that files are collected.
--- a/news/823.new.rst
+++ b/news/823.new.rst
@@ -1,2 +1,4 @@
 Add hook for ``numbers_parser`` to ensure that package's data file is
 collected.
+
+Add hook for ``sv_ttk`` to ensure that tcl files are collected.
--- a/news/823.new.rst
+++ b/news/823.new.rst
@@ -1,4 +1,2 @@
 Add hook for ``numbers_parser`` to ensure that package's data file is
 collected.
-
-Add hook for ``sv_ttk`` to ensure that files are collected.
--- a/requirements-test-libraries.txt
+++ b/requirements-test-libraries.txt
@@ -105,7 +105,7 @@ shotgun_api3==3.5.1
 slixmpp==1.8.5
 spacy==3.8.2; python_version >= '3.10'
 srsly==2.4.8
-sv_ttk==2.6.0; python_version >= '3.8'
+sv-ttk==2.6.0; python_version >= '3.8'
 swagger-spec-validator==3.0.4
 tableauhyperapi==0.0.20027
 thinc==9.1.1; python_version >= '3.9'
--- a/requirements-test-libraries.txt
+++ b/requirements-test-libraries.txt
@@ -105,7 +105,7 @@ shotgun_api3==3.5.1
 slixmpp==1.8.5
 spacy==3.8.2; python_version >= '3.10'
 srsly==2.4.8
-sv-ttk==2.6.0; python_version >= '3.8'
+sv-ttk==2.6.0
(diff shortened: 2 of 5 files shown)


rokm:
Hmm, looks like the `tkinter` is (still) broken in earlier python builds for x86_64 macOS. Can you add a decorator like this to the test to make sure it is skipped if `tkinter` is not importable?

https://github.com/pyinstaller/pyinstaller/blob/0779daae400daa4066f91dd634cc731608b901ed/tests/functional/test_hooks/test_pil.py#L27

requirements-test-libraries.txt
```diff
@@ -105,6 +105,7 @@ shotgun_api3==3.5.1
 slixmpp==1.8.5
 spacy==3.8.2; python_version >= '3.10'
 srsly==2.4.8
+sv-ttk==2.6.0; python_version >= '3.8'
```
rokm, line 108: There's no need for `; python_version >= '3.8'` since python 3.8 is the lowest version we support.

aarons20:
> Hmm, looks like the `tkinter` is (still) broken in earlier python builds for x86_64 macOS. Can you add a decorator like this to the test to make sure it is skipped if `tkinter` is not importable?
> 
> https://github.com/pyinstaller/pyinstaller/blob/0779daae400daa4066f91dd634cc731608b901ed/tests/functional/test_hooks/test_pil.py#L27

Thank you for the hint. I already tried to modify the CI to install tkinter for macOS 13 or something :)

news/823.new.rst
```diff
@@ -1,2 +1,4 @@
 Add hook for ``numbers_parser`` to ensure that package's data file is
 collected.
+
+Add hook for ``sv_ttk`` to ensure that files are collected.
```
rokm, line 4: Sorry, missed this one earlier. The news entry for this PR should go into a separate file, `news/826.new.rst`.

aarons20: Ups. Now I see where I need the pull request id. Will be changed :)

rokm:
Looks like the extra decorator did not help - I'll take a closer look tomorrow to see what exactly is going on.

aarons20:
> > Hmm, looks like the `tkinter` is (still) broken in earlier python builds for x86_64 macOS. Can you add a decorator like this to the test to make sure it is skipped if `tkinter` is not importable?
> > https://github.com/pyinstaller/pyinstaller/blob/0779daae400daa4066f91dd634cc731608b901ed/tests/functional/test_hooks/test_pil.py#L27
> 
> Thank you for the hint. I already tried to modify the CI to install tkinter for macOS 13 or something :)

Seems still to not work. I've read something about a potential tk version mismatch (8.5 vs 8.6). Perhaps that is why the decorator does not help?

aarons20:
> Looks like the extra decorator did not help - I'll take a closer look tomorrow to see what exactly is going on.

Thank you. Have a good night!

rokm:
Aha, I knew that I have previously worked around this `tk.h version (8.5) doesn't match libtk.a version (8.6)` issue in *some* test in the main repository... 

The problem is that this error is not raised on `tkinter` import, but rather when window is created (which in the case of your test happens when `set_theme` is called). So instead of decorator based on `can_import_module("tkinter")`, we will need this:

https://github.com/pyinstaller/pyinstaller/blob/0779daae400daa4066f91dd634cc731608b901ed/tests/functional/test_libraries.py#L55-L81

aarons20:
> Aha, I knew that I have previously worked around this `tk.h version (8.5) doesn't match libtk.a version (8.6)` issue in _some_ test in the main repository...
> 
> The problem is that this error is not raised on `tkinter` import, but rather when window is created (which in the case of your test happens when `set_theme` is called). So instead of decorator based on `can_import_module("tkinter")`, we will need this:
> 
> https://github.com/pyinstaller/pyinstaller/blob/0779daae400daa4066f91dd634cc731608b901ed/tests/functional/test_libraries.py#L55-L81

So since this function is in an other repository I will just copy the snippet and change the decorator?

rokm (approved):

rokm:
Thank you!

PR #832, opened 2024-11-21
Add hook-atomics.py

nmichlo:
Hook for the `atomics` library


bwoodsend:
There are more hoops to jump through than just adding the hook (with an out of date copyright header). See the [summary checklist](https://github.com/pyinstaller/pyinstaller-hooks-contrib?tab=readme-ov-file#summary).

PR #836, opened 2024-12-03
hooks: torch: suppress generation of symlinks for shared libs

### LINKED ISSUE #834: torch libtorch_cuda_linalg.so
**Describe the bug**

Missing `libtorch_cuda_linalg.so` at runtime.

**To Reproduce**

I don't have a minimal reproduction, sorry.

Error:
```
An error occurred during training.
  File "manufacia/training/pipeline.py", line 135, in run
  File "manufacia/training/patch_distribution_pipeline.py", line 354, in training_process_pipeline
  File "manufacia/training/patch_distribution_pipeline.py", line 167, in _run_training
  File "image_anomalies/patch_distribution.py", line 218, in train
  File "image_anomalies/patch_distribution.py", line 137, in patch_distribution_stats_online
RuntimeError: Error in dlopen: libtorch_cuda_linalg.so: cannot open shared object file: No such file or directory
```

**Expected behavior**

pyinstaller does correctly collect the object file `_internal/torch/lib/libtorch_cuda_linalg.so`.

I think that `pyinstaller` should symlink this shared object file in the `_internal` directory like this, but it doesn't:

```
libtorch_cuda_linalg.so -> torch/lib/libtorch_cuda_linalg.so*
```

Other libtorch object files are symlinked in the `_internal` directory:

```
libtorch_cpu.so -> torch/lib/libtorch_cpu.so*
libtorch_cuda_cpp.so -> torch/lib/libtorch_cuda_cpp.so*
libtorch_cuda_cu.so -> torch/lib/libtorch_cuda_cu.so*
libtorch_cuda.so -> torch/lib/libtorch_cuda.so*
libtorch_python.so -> torch/lib/libtorch_python.so*
libtorch.so -> torch/lib/libtorch.so*
```

 - OS: Ubuntu
 - Python Version: 3.10
 - Version of `pyinstaller-hooks-contrib`: 2024.09
 - Version of PyInstaller 6.11.0

rokm:
On Linux, have the `torch` hook use the`bindepend_symlink_suppression` hook attribute to prevent PyInstaller's binary dependency analysis from creating symlinks to top-level application directory for shared libraries discovered in the `torch/lib` directory during binary dependency analysis.

These symbolic links confuse certain `torch` builds about the location of  `torch`, and prevent `torch/lib/libtorch_cuda_linalg.so` from being (dynamically) loaded. The problematic builds are wheels provided by https://download.pytorch.org/whl/torch for `torch` 1.13.x, 2.0.x, and 2.1.x (tested with 1.13.1+cu117, 2.0.1+cu117, and 2.1.2+cu118). Later versions do not seem to be affected. The wheels provided on PyPI do not seem to be affected either, regardless of the version.

However, these symlinks should be not necessary on linux in general, so there should be no harm in suppressing them for all versions.

Fixes #834.

### DIFF
--- a/_pyinstaller_hooks_contrib/stdhooks/hook-torch.py
+++ b/_pyinstaller_hooks_contrib/stdhooks/hook-torch.py
@@ -73,6 +73,20 @@ if is_module_satisfies("PyInstaller >= 6.0"):
         logger.info("hook-torch: inferred hidden imports for CUDA libraries: %r", nvidia_hiddenimports)
         hiddenimports += nvidia_hiddenimports
 
+        # On Linux, prevent binary dependency analysis from generating symbolic links for libraries from `torch/lib` to
+        # the top-level application directory. These symbolic links seem to confuse `torch` about location of its shared
+        # libraries (likely because code in one of the libraries looks up the library file's location, but does not
+        # fully resolve it), and prevent it from finding dynamically-loaded libraries in `torch/lib` directory, such as
+        # `torch/lib/libtorch_cuda_linalg.so`. The issue was observed with earlier versions of `torch` builds provided
+        # by https://download.pytorch.org/whl/torch, specifically 1.13.1+cu117, 2.0.1+cu117, and 2.1.2+cu118; later
+        # versions do not seem to be affected. The wheels provided on PyPI do not seem to be affected, either, even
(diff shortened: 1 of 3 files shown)


bwoodsend (approved):

jamesdbrock:
This works, thank you very much.

PR #853, opened 2025-01-14
hooks: add hook for intake

mo-dkrz:
Add hook for ``intake`` to ensure that package's data and binaries of the drivers plugin are collected.

### DIFF
--- a/.github/workflows/pr-test.yml
+++ b/.github/workflows/pr-test.yml
@@ -160,6 +160,9 @@ jobs:
           # test_xarray_chunk requires dask in addition to xarray
           pip show -qq xarray && pip install --prefer-binary dask
 
+          # Install additional intake plugins for test_intake_plugins
+          pip show -qq intake && pip install --prefer-binary intake-xarray
+
           # Install PyInstaller
           pip install ${{ matrix.pyinstaller }}
 
--- a/.github/workflows/pr-test.yml
+++ b/.github/workflows/pr-test.yml
@@ -164,7 +164,7 @@ jobs:
           pip install ${{ matrix.pyinstaller }}
 
       - name: Run tests
-        run: pytest -v
+        run: pytest -v -s
 
       # Conditionally enable slow tests, so that they are ran only if
       # their corresponding packages are explicitly installed but not
--- a/.github/workflows/pr-test.yml
+++ b/.github/workflows/pr-test.yml
@@ -164,7 +164,7 @@ jobs:
           pip install ${{ matrix.pyinstaller }}
 
       - name: Run tests
-        run: pytest -v -s
+        run: pytest -v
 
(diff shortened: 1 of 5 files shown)


rokm:
In general, please avoid using `collect_all` - instead, you should use `collect_submodules`, `collect_data_files` and/or `collect_dynamic_libs` (plus `copy_metadata`) as required.

Do any of these plugins actually contain binaries and/or data files? The couple from the start of the list that I've looked at all seem to be pure-python packages without binaries and/or data.

_pyinstaller_hooks_contrib/stdhooks/hook-intake.py
```diff
@@ -0,0 +1,105 @@
...
+    'intake-dremio': {
+        'import_name': 'intake_dremio'
+    },
+    'intake-duckdb': {
+        'import_name': 'intake_duckdb'
+    }
+}
+
+for plugin_info in OPTIONAL_PLUGINS.values():
+    if is_module_satisfies(plugin_info['import_name']):
```
rokm, line 102: `is_module_satisfies` operates on dist names, so here you are (ab)using the fact that the import names happen match normalized dist names. Which makes the whole dictionary-of-dictionaries redundant - instead, it could be a simple list of import names.

mo-dkrz: done

.github/workflows/pr-test.yml
```diff
@@ -164,7 +164,7 @@ jobs:
           pip install ${{ matrix.pyinstaller }}
 
       - name: Run tests
-        run: pytest -v
+        run: pytest -v -s
```
rokm, line 167: This should be reverted.

mo-dkrz: yes, definitely

rokm:
Hmm, actually, it looks like `intake` discovers its plugins via `intake.drivers` entry points. Does your use-case still work if you simplify the hook into:
```python
# hook-intake.py
from PyInstaller.utils.hooks import collect_entry_point
datas, hiddenimports = collect_entry_point('intake.drivers')
```

mo-dkrz:
> Hmm, actually, it looks like `intake` discovers its plugins via `intake.drivers` entry points. Does your use-case still work if you simplify the hook into:
> 
> ```python
> # hook-intake.py
> from PyInstaller.utils.hooks import collect_entry_point
> datas, hiddenimports = collect_entry_point('intake.drivers')
> ```

it was completely to-the-point, yes it works perfectly on my use case and rationally speaking it has to work for other cases as well since as you correctly said `intake` discovers them through the `driver` entrypoint. This could simplify the hook a lot. I will commit the change req in a moment

mo-dkrz:
done

rokm (approved):

PR #854, opened 2025-01-14
hook: add hook for zarr

mo-dkrz:
Add hook for `zarr` to ensure that package's metadata and submodules of the drivers plugin are collected.


mo-dkrz:
i have to close this one since i forgot to rebase

PR #855, opened 2025-01-14
hooks: add hook for zarr

mo-dkrz:
Add hook for `zarr` to ensure that package's metadata and submodules of the drivers plugin are collected.


rokm:
At least on linux, the test program seems to run successfully even without the hook. Why do you need the metadata collected, and what issue are you trying to solve with the forced submodule collection?

mo-dkrz:
Thanks for your consideration! `intake-xarray` plugin of `intake` uses the the `zarr` lib which for this purpose, we need to provide the metadata of `zarr`, otherwise we get:

```bash
  File "intake.py", line 358, in _get_urlpath_metadata
  File "asyncio/threads.py", line 25, in to_thread
  File "concurrent/futures/thread.py", line 59, in run
  File "metadata_crawler/backends/intake.py", line 353, in _open_dataset
  File "xarray/backends/api.py", line 679, in open_dataset
  File "xarray/backends/zarr.py", line 1564, in open_dataset
  File "xarray/backends/zarr.py", line 703, in open_group
  File "xarray/backends/zarr.py", line 1744, in _get_open_params
  File "xarray/backends/zarr.py", line 112, in _zarr_v3
  File "xarray/namedarray/utils.py", line 56, in module_available
  File "importlib/metadata/__init__.py", line 987, in version
  File "importlib/metadata/__init__.py", line 960, in distribution
  File "importlib/metadata/__init__.py", line 409, in from_name
importlib.metadata.PackageNotFoundError: No package metadata was found for zarr
```
we can resolve this issue via `--collect-all` flag in PyInstaller, but it kinda increases the size of output binary. The same reason goes for `fsspec` as well!

Do you have any idea or approach on this?

rokm:
Thanks for the explanation. 

Strictly speaking, because `zarr` per-se does not require its own metadata., this should be accommodated by hook for `xarray.backends.zarr` (`hook-xarray.backends.zarr.py`) that collects metadata for `zarr`. 

What about hidden imports? Are there any modules actually missing?

I suppose to properly test this, the test would need to try to open a `zarr`-based dataset via `xarray`?

rokm:
> Thanks for the explanation.
> 
> Strictly speaking, because `zarr` per-se does not require its own metadata., this should be accommodated by hook for `xarray.backends.zarr` (`hook-xarray.backends.zarr.py`) that collects metadata for `zarr`.
> 
> What about hidden imports? Are there any modules actually missing?
> 
> I suppose to properly test this, the test would need to try to open a `zarr`-based dataset via `xarray`?

Or perhaps a variant of existing `test_xarray` that tries to convert data array to zarr-backed array should suffice:

```python
import xarray as xr
import numpy as np

data = xr.DataArray(
    np.random.randn(2, 3),
    dims=("x", "y"),
    coords={"x": [10, 20]},
)

data_zarr = data.to_zarr()
print(data_zarr)
```

mo-dkrz:
> What about hidden imports? Are there any modules actually missing?

No i don't think it's necessary at all. I took a test on `hook-zarr.py` via only `copy_metadata` and didn't use `hook-fsspec.py`, and got the following: 


```bash
  File "intake.py", line 358, in _get_urlpath_metadata
  File "asyncio/threads.py", line 25, in to_thread
  File "concurrent/futures/thread.py", line 59, in run
  File "metadata_crawler/backends/intake.py", line 353, in _open_dataset
  File "xarray/backends/api.py", line 679, in open_dataset
  File "xarray/backends/zarr.py", line 1564, in open_dataset
  File "xarray/backends/zarr.py", line 703, in open_group
  File "xarray/backends/zarr.py", line 1761, in _get_open_params
  File "zarr/api/synchronous.py", line 212, in open_consolidated
  File "zarr/core/sync.py", line 142, in sync
  File "zarr/core/sync.py", line 98, in _runner
  File "zarr/api/asynchronous.py", line 346, in open_consolidated
  File "zarr/api/asynchronous.py", line 800, in open_group
  File "zarr/storage/_common.py", line 305, in make_store_path
  File "zarr/storage/_fsspec.py", line 174, in from_url
  File "fsspec/core.py", line 403, in url_to_fs
  File "fsspec/core.py", line 351, in _un_chain
  File "fsspec/registry.py", line 248, in get_filesystem_class
ImportError: HTTPFileSystem requires "requests" and "aiohttp" to be installed
```

But based on my latest trial, the solution for jumping over this issue, is to use the submitted `hook-fsspec` strategy without any touch and using only `copy_metadata('zarr')` in `zarr-hook` without hidden-import

mo-dkrz:
> Or perhaps a variant of existing `test_xarray` that tries to convert data array to zarr-backed array should suffice:

smart idea, I can do it, i will commit in a moment

_pyinstaller_hooks_contrib/stdhooks/hook-zarr.py
```diff
@@ -0,0 +1,15 @@
+# ------------------------------------------------------------------
+# Copyright (c) 2024 PyInstaller Development Team.
```
rokm, line 2: 2025

news/855.new.rst
```diff
@@ -0,0 +1,2 @@
+Add hook for ``zarr`` to ensure that package's metadata and
```
rokm, line 1: The hook is collecting only metadata.

tests/test_libraries.py
```diff
@@ -2078,6 +2078,7 @@ def test_xarray(pyi_builder):
             dims=("x", "y"),
             coords={"x": [10, 20]},
         )
+        data_zarr = data.to_zarr()
```
rokm, line 2081: Ah, that's not what I had I mind. I meant to create a copy of `test_xarray` (named `test_xarray_to_zarr`), which has the added conversion to `zarr`. And this new test requires both `@importorskip('xarray')` and `@importorskip('zarr')`.

rokm (changes requested):

news/855.new.rst
```diff
@@ -0,0 +1 @@
+The hook is collecting only metadata.
```
rokm, line 1: Sorry, I wasn't clear enough - the news fragment should be: 
```
Add hook for ``zarr`` to collect the package's metadata.
```

requirements-test-libraries.txt
```diff
@@ -238,6 +238,7 @@ pysaml2==7.5.0; python_version >= '3.9'
 pysaml2==7.3.0; python_version < '3.9'  # pyup: ignore
 toga==0.4.8; python_version >= '3.9'
 numbers-parser==4.14.2; python_version >= '3.9'
+zarr==3.0.0; python_version >= '3.9'
```
rokm, line 241: Looks like we'll need `python_version >= '3.11'` here.

mo-dkrz:
do i need to downgrade the version of zarr since the newest version seems to not be compatible with py3.9? or change the `python_version >= '3.9'` to `python_version >= '3.11'`?

mo-dkrz:
@bwoodsend conflicts are resolved for `hook-zarr` now, after merging this one i go for `hook-fsspec`

PR #857, opened 2025-01-17
hooks: update `numba` hook for compatibility with upcoming v0.61

rokm:
Add hidden imports for dynamically redirected and loaded modules that were introduced as part of the new type system in upcoming `numba` v0.61.0.

Extend `numba` tests with a basic test that uses `numba.jit`, and a basic import test with `NUMBA_USE_LEGACY_TYPE_SYSTEM` set to 0.

Closes pyinstaller/pyinstaller#8989.

Oneshot run (0.60, 0.61rc1, 0.61rc2): https://github.com/rokm/pyinstaller-hooks-contrib/actions/runs/12828775609

### DIFF
--- a/_pyinstaller_hooks_contrib/stdhooks/hook-numba.py
+++ b/_pyinstaller_hooks_contrib/stdhooks/hook-numba.py
@@ -25,3 +25,30 @@ hiddenimports = ["llvmlite"]
 # around for backward compatibility with existing pickled data, but does not import it directly anymore.
 if is_module_satisfies("numba >= 0.59.0"):
     hiddenimports += ["numba.cloudpickle.cloudpickle_fast"]
+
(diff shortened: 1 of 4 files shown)


rokm:
If these new/old modules turn to be wildly changing between releases, we can switch to a `collect_submodules` as suggested in https://github.com/pyinstaller/pyinstaller/issues/8989#issuecomment-2597115458; but I would prefer to stick to a fixed list if necessary.

bwoodsend (commented):
I think I'd just go for `collect_submodules()` at this point but it's up to you...

bwoodsend (approved):

rokm:
> I think I'd just go for `collect_submodules()` at this point but it's up to you...

Let's first see how this evolves through next couple of releases. At the moment, fixed list seems manageable.

PR #862, opened 2025-01-31
Deprecated-hook

mo-dkrz:
Add hook for `deprecated` to ensure that submodules of the lib are collected.


bwoodsend:
What makes you think this is needed? I can't see any signs of `deprecated` doing anything that PyInstaller wouldn't be able to detect.

mo-dkrz:
> What makes you think this is needed? I can't see any signs of `deprecated` doing anything that PyInstaller wouldn't be able to detect.

@bwoodsend Thanks for the response. I was bundling a library which was zarr-included, so it was complaining about the `deprecated` modules, so by adding `deprecated` as hidden-import it was fixed. I thought it would be nice to fix it fundamentally that might fix the issue of others as well. Or another solution would be updating the `zarr-hook` and collect the modules there!

```sh
  File "PyInstaller/loader/pyimod02_importers.py", line 384, in exec_module
  File "zarr/__init__.py", line 2, in <module>
  File "PyInstaller/loader/pyimod02_importers.py", line 384, in exec_module
  File "zarr/codecs.py", line 2, in <module>
  File "PyInstaller/loader/pyimod02_importers.py", line 384, in exec_module
  File "numcodecs/__init__.py", line 43, in <module>
  File "numcodecs/blosc.pyx", line 9, in init numcodecs.blosc
ModuleNotFoundError: No module named 'deprecated'
```

rokm:
Adding a hook for `deprecated` won't help you in any way in this case; you need a hook for the module that has this hidden import (in this case, `numcodecs.blosc` cythonized extension).

This should be already fixed by https://github.com/pyinstaller/pyinstaller-hooks-contrib/pull/858, but we should probably release a new version that contains that change.

PR #881, opened 2025-03-18
hooks: add hook for python-dateutil

lambda:
This contains a data file dateutil-zoneinfo.tar.gz that we need to include for full functionality.

### DIFF
--- a/requirements-test-libraries.txt
+++ b/requirements-test-libraries.txt
@@ -29,7 +29,7 @@ dash==2.18.2
 dash-bootstrap-components==1.7.1; python_version >= '3.9'
 dash-uploader==0.6.1
 dask[diagnostics,array,distributed]==2025.2.0; python_version >= '3.10'
-python-dateutil=2.9.0.post0
+python-dateutil==2.9.0.post0
 # discid requires libdiscid to be provided by the system.
 # We install it via apt-get and brew on ubuntu and macOS CI runners, respectively.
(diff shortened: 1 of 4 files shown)


requirements-test-libraries.txt
```diff
@@ -29,6 +29,7 @@ dash==2.18.2
 dash-bootstrap-components==1.7.1; python_version >= '3.9'
 dash-uploader==0.6.1
 dask[diagnostics,array,distributed]==2025.2.0; python_version >= '3.10'
+python-dateutil=2.9.0.post0
```
rokm, line 32: This should be `==`.

tests/test_libraries.py
```diff
@@ -2623,3 +2623,11 @@ def test_pypdfium2(pyi_builder):
     pyi_builder.test_source("""
         import pypdfium2
     """)
+
```
rokm, line 2626: The functions need to be separated by two empty lines to appease the linter.

rokm (changes requested):

requirements-test-libraries.txt
lambda: Whoops, fixed.

tests/test_libraries.py
lambda: Also fixed, thanks for the review!

rokm (approved):

bwoodsend:
Thank you!

Issue #873, opened 2025-03-05
Assistance with Debian packaging


sorenstoutner:
Now that pyinstaller has been accepted into Debian I am working on packaging pyinstaller-hooks-contrib.  I have a basic package prepared, but it currently skips a lot of tests.  Some of them require components that are not currently in Debian, which I am unlikely to worry about at this time.  But others I think can be addressed.

Some of these I have been able to figure out myself, but there are a few where you might be able to help me.  My intention is to ask one question at a time.

When I include altair in the build environment, I receive the following error:

`RecursionError: maximum recursion depth exceeded`

https://salsa.debian.org/soren/pyinstaller-hooks-contrib/-/jobs/7201942#L6013

The python3-altair package in included in the build environment here:

https://salsa.debian.org/soren/pyinstaller-hooks-contrib/-/blob/debian/master/debian/control?ref_type=heads#L13

The version of python-altair in Debian is 5.0.2, which is a bit behind the upstream 5.5.0.

https://tracker.debian.org/pkg/python-altair

If you think the version is the problem I can work on updating python-altair, but this error message doesn't immediately point me toward a version problem.

For reference, builds that don't include python3-altair complete successfully.

https://salsa.debian.org/soren/pyinstaller-hooks-contrib/-/pipelines/826946

bwoodsend:
Hmm, this is going to be tricky.

First warning is that `requirements-test-libraries.txt` is so large that we can't come close to installing the whole thing into one environment without dependency version conflicts (I'm guessing Debian removes at least some of more arbitrary constraints so you might manage more than we do?). We do all our testing incrementally, only installing and testing the subset of packages that have released a new version.

Second warning is that PyInstaller does struggle when environments become insanely large where the module->module dependency tree gets so deep that the stack depth exceeds the recursion limit. It's worst in packages that have large numbers of (usually optional or interchangeable) dependencies. You could probably get `altair` in if you pulled enough other packages out. Or we do implicitly raise the recursion limit for known bad offenders and I wouldn't mind doing the same for others.
https://github.com/pyinstaller/pyinstaller-hooks-contrib/blob/6470ebd203a7f6436c5a85ba164af59006fd370b/_pyinstaller_hooks_contrib/stdhooks/hook-langchain.py#L19-L25

Large environments do also slow the builds down. I don't know how much you consider that to be a factor when deciding how many libraries to include?

sorenstoutner:
[...]

Issue #883, opened 2025-03-19
Hook for pyogrio


MaximeROUBY:
**Which library is the hook for?**

`pyogrio`

**Have you gotten the library to work with pyinstaller?**

Yes, but it needs several hidden imports.

**Additional context**

If I do this it works : 

```
from PyInstaller.utils.hooks import collect_submodules

hiddenimports = collect_submodules('pyogrio')
```

Issue #882, opened 2025-03-19
Hook for rasterio


MaximeROUBY:
<!-- NO IMAGES --!>

**Which library is the hook for?**

`rasterio`

**Have you gotten the library to work with pyinstaller?**

Yes, but it needs several hidden imports.

**Additional context**

If I do this code, it works : 

```
from PyInstaller.utils.hooks import collect_submodules

hiddenimports = collect_submodules('rasterio')
```

PR #886, opened 2025-03-21
Debian fixes

bwoodsend:
Assorted fixes from #873. The first two are in response to custom Debian repackaging. I don't understand the third – it looks like the `pyqtgraph` one should never have worked before.

[Oneshot](https://github.com/bwoodsend/pyinstaller-hooks-contrib/actions/runs/13981821570) + [pyqtgraph on Linux](https://github.com/bwoodsend/pyinstaller-hooks-contrib/actions/runs/13982044546) (needs xvfb)

### DIFF
--- a/_pyinstaller_hooks_contrib/rthooks/pyi_rth_pyqtgraph_multiprocess.py
+++ b/_pyinstaller_hooks_contrib/rthooks/pyi_rth_pyqtgraph_multiprocess.py
@@ -26,9 +26,9 @@ def _setup_pyqtgraph_multiprocess_hook():
     if len(sys.argv) == 2 and sys.argv[1] == os.path.join(sys._MEIPASS, 'pyqtgraph', 'multiprocess', 'bootstrap.py'):
         # Load as module; this requires --hiddenimport pyqtgraph.multiprocess.bootstrap
         try:
-            mod_name = 'pyqtgraph.multiprocess.bootstrap'
-            mod = __import__(mod_name)
-            bootstrap_co = mod.__loader__.get_code(mod_name)
+            import importlib.util
+            spec = importlib.util.find_spec("pyqtgraph.multiprocess.bootstrap")
+            bootstrap_co = spec.loader.get_code("pyqtgraph.multiprocess.bootstrap")
         except Exception:
             bootstrap_co = None
 
@@ -37,6 +37,7 @@ def _setup_pyqtgraph_multiprocess_hook():
             sys.exit(0)
 
         # Load from file; requires pyqtgraph/multiprocess/bootstrap.py collected as data file
+        # This is obsolete for PyInstaller >= v6.10.0
         bootstrap_file = os.path.join(sys._MEIPASS, 'pyqtgraph', 'multiprocess', 'bootstrap.py')
         if os.path.isfile(bootstrap_file):
             with open(bootstrap_file, 'r') as fp:
--- a/_pyinstaller_hooks_contrib/stdhooks/hook-h5py.py
+++ b/_pyinstaller_hooks_contrib/stdhooks/hook-h5py.py
@@ -13,4 +13,6 @@
 Hook for http://pypi.python.org/pypi/h5py/
 """
 
-hiddenimports = ['h5py._proxy', 'h5py.utils', 'h5py.defs', 'h5py.h5ac']
+from PyInstaller.utils.hooks import collect_submodules
+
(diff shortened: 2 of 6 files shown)


rokm:
`pyqtgraph` worked before pyinstaller/pyinstaller#8695 and pyinstaller/pyinstaller#8934, i.e., when we had a monolithic `PyiFrozenImporter`; with that implementation, one could load *any* frozen module, take the loader from it, and use that loader to fetch the code for any other frozen module. Which is why the issue with module mismatch went unnoticed back then - and since importer refactoring, `pyqtgraph` did not have a new release.

Now that I look at that run-time hook a bit closer (and after some time since I wrote it), I think `__import__()` is the wrong way to go about things there, because it will execute the module. This is essentially a no-op, because the code has a proper `if __name__ == '__main__'` guard, but is still ugly. Instead, we should use `importlib.util.find_spec()` to find the spec, and take the loader from there.

Similarly, the fallback "load from file" block is reduntant - because if the .py file exists at the expected location, the module loading / spec look up will discover it as well (albeit using the python's built in finder and loader, rather than our own - perhaps that did not work earlier due to module name mismatch, but it should now...).

rokm:
In longer term (i.e., beyond the scope of this PR), we should probably refactor `pyproj` hook to first try to collect vendored data, and then fall-back to collecting system-wide copy, instead of trying to do both. 

Because there might be proj (and its data) installed on the system while `pyproj` wheel is installed in the python environment, and we wouldn't want to collect both vendored and system-wide data. Right now, this would likely happen only if `pyproj` wheel was installed in the base python environment (because [this check](https://github.com/pyinstaller/pyinstaller-hooks-contrib/blob/960af41fae0175bc6a6877245806e0f3b575fe53/_pyinstaller_hooks_contrib/stdhooks/hook-pyproj.py#L38-L41) does not properly account for `venv` virtual environments; so using `dnf`-installed `pyproj` in virtual environment created with `python -m venv --system-site-packages` does not work) or if it was installed inside Anaconda environment that happened to have `proj` (but not `pyproj`) installed via `conda`.

Also, I see little point in collecting system-wide data into `Library\share\proj` (Windows) and `share/proj` (non-Windows) when our run-time hook needs to set `PROJ_PATH` to point to that location; it would probably make more sense to collect into vendored directory, `pyproj/proj_dir` and have `PROJ_PATH` point to that.

That said, `rasterio` and `pyogrio` from #882 and #883 also seem to vendor copies of `proj_data` and `gdal_data`, so perhaps it would make sense to have hooks for all these packages collect the data into top level `proj_data` and `gdal_data` directories to avoid unnecessary duplication (assuming the data is, in fact, identical / compatible across the packages - which might also not be the case).

rokm (approved):

bwoodsend:
> Now that I look at that run-time hook a bit closer (and after some time since I wrote it), I think __import__() is the wrong way to go about things there, because it will execute the module.

Running `importlib.util.find_spec("pyqtgraph.multiprocess.bootstrap")` simplicity imports `pyqtgraph` anyway so I don't think we're doing any avoidable imports?

rokm:
> > Now that I look at that run-time hook a bit closer (and after some time since I wrote it), I think **import**() is the wrong way to go about things there, because it will execute the module.
> 
> Running `importlib.util.find_spec("pyqtgraph.multiprocess.bootstrap")` simplicity imports `pyqtgraph` anyway so I don't think we're doing any avoidable imports?

When we do `__import__('pyqtgraph.multiprocess.bootstrap')`, it imports and executes the module `pyqtgraph.multiprocess.bootstrap` - whereas what we really want here is just to retrieve its code object, and then execute it as if it were a main script (because that is the behavior we are emulating), i.e., with `__name__` set to `__main__`. Which is what the `exec()` part does. 

With current implementation, we end up executing the code twice - first time on import with `__name__ = 'pyqtgraph.multiprocess.bootstrap'` (which is mostly no-op, due to `if __name__ == '__main__'` guard in the file), and second time with `__name__ =  '__main__'`. This is the part that could be avoided - by using `find_spec` to look-up the loader and retrieve the code object, without explicitly importing the module (which is not really supposed to be a module, but a script that we collect as a module for our convenience).

bwoodsend:
Ahh, I see. Importing `pyqtgraph` in unavoidable but importing `pyqtgraph.multiprocess.bootstrap` is not.

rokm (approved):

PR #887, opened 2025-03-21
Replace use of get_installer with backported get_installer_for_dist

rokm:
Backport `get_installer_for_dist` hook utility function from pyinstaller/pyinstaller#9067 into `_pyinstaller_hooks_contrib.compat`, and update hooks for `enchant` and `rtree` to use it instead of the old (and to-be deprecated) `PyInstaller.utils.hooks.get_installer`.

### DIFF
--- a/_pyinstaller_hooks_contrib/compat.py
+++ b/_pyinstaller_hooks_contrib/compat.py
@@ -40,3 +40,17 @@ else:
         # Validate the version
         if packaging.version.parse(importlib_metadata.version("importlib-metadata")) < packaging.version.parse("4.6"):
             raise ImportlibMetadataError()
+
+
+# Back-port of `PyInstaller.utils.hooks.get_installer_for_dist` from PyInstaller > 6.12, made available here for
+# compatibility with older PyInstaller versions.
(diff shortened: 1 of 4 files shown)


bwoodsend:
https://github.com/pyinstaller/pyinstaller/pull/9071 makes this unneeded, right? There's no reason to wait for this one before pressing the release button?

rokm:
> [pyinstaller/pyinstaller#9071](https://github.com/pyinstaller/pyinstaller/pull/9071) makes this unneeded, right? There's no reason to wait for this one before pressing the release button?

Indeed. With https://github.com/pyinstaller/pyinstaller/pull/9071, this PR can be closed.

I'll open one later to change

https://github.com/pyinstaller/pyinstaller-hooks-contrib/blob/502174bbc412029bf0836536623c3bf983491913/_pyinstaller_hooks_contrib/stdhooks/hook-enchant.py#L45

into 

```python
if is_module_satisfies('PyInstaller > 6.12.0'):
    installer = get_installer('pyenchant') 
else:
    installer = get_installer('enchant') 
```

for the sake of technical correctness, but that should not block the release.

PR #888, opened 2025-03-23
hooks: add hook for niquests

Ousret:
I am the maintainer of Niquests.
There was one demand of a particular user that prompted me to write that hook.

related to https://github.com/jawah/niquests/issues/225

### DIFF
--- a/_pyinstaller_hooks_contrib/stdhooks/hook-niquests.py
+++ b/_pyinstaller_hooks_contrib/stdhooks/hook-niquests.py
@@ -1,0 +1,20 @@
+# ------------------------------------------------------------------
+# Copyright (c) 2025 PyInstaller Development Team.
+#
+# This file is distributed under the terms of the GNU General Public
+# License (version 2.0 or later).
+#
+# The full license is available in LICENSE, distributed with
+# this software.
+#
+# SPDX-License-Identifier: GPL-2.0-or-later
+# ------------------------------------------------------------------
+
+from PyInstaller.utils.hooks import collect_submodules
+
+
+hiddenimports = [
+    *collect_submodules("niquests"),
+    *collect_submodules("urllib3"),
+    *collect_submodules("urllib3_future"),
+]
--- a/requirements-test-libraries.txt
+++ b/requirements-test-libraries.txt
@@ -249,7 +249,8 @@ zarr==3.0.5; python_version >= '3.11'
 intake==2.0.8; python_version >= '3.9'
 h3==4.2.2
 selectolax==0.3.28
- ruamel.yaml.string==0.1.1
+ruamel.yaml.string==0.1.1
+niquests==3.13.1
 
 # ------------------- Platform (OS) specifics
 
--- a/news/888.new.rst
+++ b/news/888.new.rst
@@ -1,0 +1,1 @@
+Add hook for ``niquests``.
--- a/tests/test_libraries.py
+++ b/tests/test_libraries.py
@@ -2632,3 +2632,34 @@ def test_dateutil(pyi_builder):
 
         assert getzoneinfofile_stream()
     """)
+
+
+@importorskip('niquests')
+def test_niquests(pyi_builder):
+    pyi_builder.test_source("""
+        import niquests
+
+        try:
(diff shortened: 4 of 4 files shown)


_pyinstaller_hooks_contrib/stdhooks/hook-niquests.py
```diff
@@ -0,0 +1,26 @@
...
+
+from PyInstaller.utils.hooks import collect_submodules, is_package
+
+
+hiddenimports = [
+    *collect_submodules("niquests"),
+    *collect_submodules("urllib3"),
+    *collect_submodules("urllib3_future"),
+    *collect_submodules("jh2"),
+    *collect_submodules("h11"),
```
rokm, line 21: Are these really all required? The provided test seems to be working (at least on linux) with just `urllib3.contrib.resolver.system` and `niquests.packages` added to hidden imports. And even the latter is needed only because of [these imports](https://github.com/jawah/niquests/blob/f42b2bb873e3e516ad8044d69a5c72a4f8b8c88e/src/niquests/__init__.py#L49-L55) that imply that `niquests.packages` should be a package whereas it is a module, and throw off our analysis.

Ousret: It is necessary. We can't be really precise here.

I have added complementary tests. But the main idea here, is that "urllib3-future" have many lazy loaded sub modules. not just `urllib3.contrib.resolver.system` but  `urllib3.contrib.hface.protocols.http3` or even `urllib3.contrib.webextensions`. There's plan for others in the future, so the hook is more resilient this way.
FYI: I've removed h11, jh2 and qh3 explicit collection as PyInstaller seems to find it automatically.

Regards,

rokm: Hmm, the extended test code still seems to run for me with just the afore-mentioned two hidden imports, but I get what you are saying.

Just to clarify, though, are these lazy imports done by `niquest` or are they coming from within `urllib3` / `urllib3_future` itself?

As for the PR,
1. add `niquest` to [`requirements-test-libraries.txt`](https://github.com/pyinstaller/pyinstaller-hooks-contrib/blob/master/requirements-test-libraries.txt) so that the test will be ran when the package is updated
2. add a [news fragment](https://github.com/pyinstaller/pyinstaller-hooks-contrib/blob/master/news/README.txt) for the changelog - the entry can be simply `Add hook for ``niquests``.`

Ousret: both niquests and urllib3/urllib3-future are concerned.

1. done.
2. already added, maybe you missed it.

requirements-test-libraries.txt
```diff
@@ -249,7 +249,8 @@ zarr==3.0.5; python_version >= '3.11'
 intake==2.0.8; python_version >= '3.9'
 h3==4.2.2
 selectolax==0.3.28
- ruamel.yaml.string==0.1.1
+ruamel.yaml.string==0.1.1
+niquests>=3,<4
```
rokm, line 253: We tend to pin the latest version here, because we have pyup-bot weekly update these entries and run tests for them. That way, we can see what packages need updated hooks for their new releases.

_pyinstaller_hooks_contrib/stdhooks/hook-niquests.py
```diff
@@ -0,0 +1,20 @@
+# ------------------------------------------------------------------
+# Copyright (c) 2023 PyInstaller Development Team.
```
rokm, line 2: I missed this earlier, but the year should be 2025.

requirements-test-libraries.txt
Ousret: done

_pyinstaller_hooks_contrib/stdhooks/hook-niquests.py
Ousret: done

rokm (approved):

rokm:
LGTM. Thank you!

PR #894, opened 2025-04-01
This week's chores...

bwoodsend:
Code changes from https://github.com/pyinstaller/pyinstaller-hooks-contrib/pull/892


rokm:
I guess we need `python_version >= '3.9' or (sys_platform != 'darwin' or platform_machine != 'arm64')` marker for both `apkutils` and `adbutils`, since the former depends on `lxml`, and `lxml` does not provide binary wheels for python 3.8 + macOS + arm64 combination (and building `lxml` from source results in defunct installation).

rokm (approved):

rokm:
Guess we can the marker to `adbutils` as part of #892 if (when) necessary.

PR #899, opened 2025-04-10
Add TkinterWeb and TkinterWeb-Tkhtml hooks

Andereoo:
Hello!

I've been getting complaints about my package TkinterWeb not working with PyInstaller and thought I would add a hook now that the package is more mature.

I included hooks for both TkinterWeb and its child package TkinterWeb-Tkhtml but only included a test for Tkinterweb as it also tests TkinterWeb-Tkhtml indirectly. 

Hopefully this will suffice?

Thanks!

### DIFF
--- a/_pyinstaller_hooks_contrib/stdhooks/hook-tkinterweb.py
+++ b/_pyinstaller_hooks_contrib/stdhooks/hook-tkinterweb.py
@@ -1,0 +1,16 @@
+# ------------------------------------------------------------------
+# Copyright (c) 2025 PyInstaller Development Team.
+#
+# This file is distributed under the terms of the GNU General Public
+# License (version 2.0 or later).
+#
+# The full license is available in LICENSE, distributed with
+# this software.
+#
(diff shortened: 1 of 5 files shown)


bwoodsend:
You can add one of these *if testing x, also install y* workarounds if you prefer.
https://github.com/pyinstaller/pyinstaller-hooks-contrib/blob/2ba7dbb8fa0ef4cdd55307c26381398236ba31f0/.github/workflows/pr-test.yml#L164-L165

Andereoo:
Fair enough! I just added a second test. It is probably more clear and foolproof to have a test for each hook anyway.

Andereoo:
Thank you very much!

bwoodsend:
Thanks!

One request, can you remember to set the minium supported Python version in your packages (`python_requires` in the `setup.py`) if/when you end up dropping support for Python 3.8.

Andereoo:
Gladly! It should be a while before I drop Python 3.8 support but I will
include python_requires in my next release so I don't forget! I probably
should have been doing that all along anyway; thank you!

PR #906, opened 2025-04-30
hooks: usb: fix binary collect on windows

### LINKED ISSUE #905: 'libusb0.dll' is excluded in `hook-usb` on Windows platform
The reason is that the `_resolveCtypesImports(['libusb0.dll'])` method will return empty on Windows( not test other platform):
https://github.com/pyinstaller/pyinstaller-hooks-contrib/blob/f89be6debff540da250850f91e0a67021e3f12bc/_pyinstaller_hooks_contrib/stdhooks/hook-usb.py#L52

and  _resolveCtypesImports will call:
https://github.com/pyinstaller/pyinstaller/blob/c04bea46d83ac4d5c955d2f0140df621740e9941/PyInstaller/depend/utils.py#L244

finally call:
https://github.com/pyinstaller/pyinstaller/blob/c04bea46d83ac4d5c955d2f0140df621740e9941/PyInstaller/depend/dylib.py#L327
here `check_library` will exclude everything from the Windows directory:
https://github.com/pyinstaller/pyinstaller/blob/c04bea46d83ac4d5c955d2f0140df621740e9941/PyInstaller/depend/dylib.py#L299-L312

hope for fix it

rqx110:
resolve #905

### DIFF
--- a/_pyinstaller_hooks_contrib/stdhooks/hook-usb.py
+++ b/_pyinstaller_hooks_contrib/stdhooks/hook-usb.py
@@ -41,15 +41,16 @@ if not skip_pyusb_discovery:
         usb.core.find()
         # get the backend symbols which have been added (loaded)
         backends = set(dir(usb.backend)) - backend_contents_before_discovery
-        # gather the libraries from the loaded backends
-        backend_lib_basenames = []
         for usblib in [getattr(usb.backend, be)._lib for be in backends]:
             if usblib is not None:
-                # OSX returns the full path, Linux only the filename.
-                # save the basename and reconstruct the path after gathering.
-                backend_lib_basenames.append(os.path.basename(usblib._name))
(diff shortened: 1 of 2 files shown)


rokm:
This looks reasonable. Perhaps the change should be (eventually) extended to avoid using `_resolveCtypesImports`-based resolution not just on Windows, but whenever the library name is already an absolute path (which should, for example, as be the case on macOS).

Please add a short [news fragment](https://github.com/pyinstaller/pyinstaller-hooks-contrib/tree/master?tab=readme-ov-file#add-a-news-entry) for the changelog (for example, something along the lines of ` ``usb`` hook: fix shared library collection on Windows.`), and then we should be ready to merge.

rqx110:
@rokm ready for review

rokm (approved):

Issue #912, opened 2025-05-28
usb.backend.libusb1.get_backend(find_library) doesn't work.


zapta:
**Describe the bug**

The following code uses a callback function ``find_library`` to allow custom usb backend loading. It works under standard python but not under pyinstaller.
```
backend = usb.backend.libusb1.get_backend(find_library=find_library)
```

* Which hook/library isn't working? 

``pyi_rth_usb.py``

* Does the error get raised while building or when running?

When running.

**To Reproduce**

A minimal example file:
```
import usb.backend.libusb1

# Fake callback.
def find_library(name):
    print(f"Looking for: {name}")
    return None

# This should call the callback at least once.
backend = usb.backend.libusb1.get_backend(find_library=find_library)
```

PyInstaller command:
```
pyinstall test.py
```

Error:
```
$ ./dist/test/test 
$
```

**Expected behavior**
Callback should be called and print messages.
```
$ python test.py 
Looking for: usb-1.0
Looking for: libusb-1.0
Looking for: usb
$
```

**Screenshots**
N/A

**Desktop (please complete the following information):**
 - OS:  Mac OSX Apple Silicon 
 - Python Version:  Python 3.13.1
 - Version of `pyinstaller-hooks-contrib`:   2025.4
 - Version of PyInstaller   6.13.0

**Additional context**
The usb runtime hook ignores the ``find_library`` argument which causes loading using ``find_library`` callback to fail. When ``find_library`` is not passed, the hook should use it to find the lib path instead of looking in the bundle.

https://github.com/pyinstaller/pyinstaller-hooks-contrib/blob/3cc7a3a3df4a7b404ecf3fedfbcbe48980bd61f8/_pyinstaller_hooks_contrib/rthooks/pyi_rth_usb.py#L30

PR #913, opened 2025-05-28
Fixed issue #912 (USB find_library callback)

zapta:
With this change, the ``find_library`` argument is now handled properly which fixed custom libusb file loading under pyinstaller. See issue #912 for details.


rokm (approved):
Looks reasonable to me. Thanks!

PR #922, opened 2025-07-09
Add hook for pynng

johnthagen:
On a PyInstaller application that had `pynng` as a dependency, this error was thrown when the executable started up:

```
  File "PyInstaller/loader/pyimod02_importers.py", line 457, in exec_module
  File "pynng/__init__.py", line 3, in <module>
ModuleNotFoundError: No module named '_cffi_backend'
```

`cffi` is a dependency of `pynng`:

https://github.com/codypiersall/pynng/blob/2179328f8a858bbb3e177f66ac132bde4a5aa859/pyproject.toml#L22

I based this PR off of a similar one for `bcrypt`

- https://github.com/pyinstaller/pyinstaller/pull/4750/

### DIFF
--- a/requirements-test-libraries.txt
+++ b/requirements-test-libraries.txt
@@ -96,6 +96,7 @@ pyqtgraph==0.13.7; python_version >= "3.9"
 pyusb==1.3.1; python_version >= "3.9"
 pyviz-comms==3.0.6
 pyvjoy==1.0.1; sys_platform == "win32"
+pynng==0.8.1; python_version >= "3.7"
 pynput==1.8.1
 # pymssql provides only x86_64 macOS wheels for python 3.9 and 3.10. But at the time of writing (v2.3.2), the universal2 wheels are broken on arm64 macOS as well.
 pymssql==2.3.6; python_version >= "3.9" and (sys_platform != "darwin" or platform_machine != "arm64")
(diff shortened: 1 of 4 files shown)


requirements-test-libraries.txt
```diff
@@ -96,6 +96,7 @@ pyqtgraph==0.13.7; python_version >= "3.9"
 pyusb==1.3.1; python_version >= "3.9"
 pyviz-comms==3.0.6
 pyvjoy==1.0.1; sys_platform == "win32"
+pynng==0.8.1; python_version >= "3.7"
```
rokm, line 99: We support only python >= 3.8 here, so `; python_version >= "3.7"` is redundant.

tests/test_libraries.py
```diff
@@ -2878,3 +2878,9 @@ def test_narwhals(pyi_builder):
     pyi_builder.test_source("""
         import narwhals
     """)
+
```
rokm, line 2881: Linter requires functions to be separated by two empty lines.

johnthagen: Beat me by a second! 😆  
4a1c752

requirements-test-libraries.txt
johnthagen: Fixed in  b1b72fa

bwoodsend (approved):

PR #925, opened 2025-07-20
Tests: Requirements: Scheduled weekly dependency update for week 29

rokm:


rokm:
This time around, there are couple of recursion limit failures coming from different tests, which might be coming from a common optional dependency chain. I think at this point, it makes sense to just raise the recursion limit as part of `_pyinstaller_hooks_contrib/__init__.py` rather than peppering the corresponding code snippet over different hooks (arguably it would make sense for PyInstaller itself to raise the limit, but if we do it here, we also cover older PyInstaller versions).

rokm:
On a semi-related note, I think our recursion-limit-error message should be modified to display the current value of the recursion limit. That way, when a user posts the message, we could immediately tell whether they successfully tried to raise the limit or not...

bwoodsend (approved):

rokm:
> > I'm also fine with moving this limit-raising from _pyinstaller_hooks_contrib/**init**.py into Analysis.assemble (so that it takes effect after potential sys.setrecursionlimit() in the .spec file), if we can live with a) this being available only from next PyInstaller release on, and b) not being able to clean up the hooks (unless we're fine with breaking older PyInstaller versions).
> 
> I don't want to block (or make awkward) a user from setting their own recursion limit. Even if we're confident that 5k should be a universally acceptable value (which I'm not).

If user sets the value below 5k, then (some of) the hooks override that to 5k anyway (while preserving higher values).

What are your thoughts on monkey-patching `Analysis.assemble()` in `_pyinstaller_hooks_contrib/__init__.py`? That should have the same effect as the change currently made by hooks, because it would happen after user potentially set the value.

bwoodsend:
> What are your thoughts on monkey-patching Analysis.assemble() in _pyinstaller_hooks_contrib/__init__.py? That should have the same effect as the change currently made by hooks, because it would happen after user potentially set the value.

I don't want to couple the two projects like that. For all we know, there might not always be a `Analysis.assemble()` after the modulegraph migration or some bid to properly separate PyInstaller's public and internal APIs.

PR #931, opened 2025-07-27
hooks: update eccodes-related hooks for compatibility with eccodes 2.43.0

rokm:
The non-Windows PyPI wheels for `eccodes` 2.43.0 do not provide the bundled shared libraries anymore, but instead depend on `eccodeslib` wheels for that.

First, we need to fix the override for `findlibs.find()` in the run-time hook for `findlibs` to avoid providing the default value for `pkg_name` argument; instead, pass the original (`None`) value when calling the original implementation. The older versions of `findlibs` (e.g., 0.0.5) used `lib_name` as default for `pkg_name`, while the recent versions (>= 1.0.0) use `lib_name` + `'lib'` (i.e., `eccodes` -> `eccodeslib`).

Second, `findlibs.find()` needs to be able to import `eccodeslib` to query its `__file__` attribute, so add a hidden import to ensure that module is always collected (otherwise, `eccodeslib` might become a namespace package without `__file__` attribute).

Finally, `eccodeslib` declares transitive dependencies on `eckitlib` and `fckitlib` to `findlibs`, so `findlibs.find()` also attempts to import these packages to query their `__file__` attribute; if the packages are not importable, a `UserWarning` is emitted at run time. Add a hook for `eccodeslib` that adds hidden imports for `eckitlib` and `fckitlib`.

Just in case there is any dynamic loading going on (now or in the future), have the hooks for `eccodeslib`, `eckitlib`, and `fckitlib` collect all their bundled shared libraries.

### DIFF
--- a/_pyinstaller_hooks_contrib/rthooks/pyi_rth_findlibs.py
+++ b/_pyinstaller_hooks_contrib/rthooks/pyi_rth_findlibs.py
@@ -31,8 +31,7 @@ def _pyi_rthook():
 
     _orig_find = getattr(findlibs, 'find', None)
 
-    def _pyi_find(lib_name, pkg_name=None):
-        pkg_name = pkg_name or lib_name
+    def _pyi_find(lib_name, *args, **kwargs):
         extension = findlibs.EXTENSIONS.get(sys.platform, ".so")
 
         # First check sys._MEIPASS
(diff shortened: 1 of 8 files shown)


rokm:
Oneshot test (albeit without system-installed libraries that could interfere with the bundled ones): https://github.com/rokm/pyinstaller-hooks-contrib/actions/runs/16551750658

PR #942, opened 2025-08-31
Tests: Requirements: Scheduled weekly dependency update for week 35

rokm:


rokm:
Skipping `compliance-checker` v5.4.0 because it doesn't work when ran unfrozen, either: https://github.com/ioos/compliance-checker/issues/1232

bwoodsend (approved):

bwoodsend:
I'd quite like to make `pyi_builder.test_source()` run the unfrozen code in a subprocess either before the actual test or after if it fails, just so that we can make these upstream breakages easier to triage.

rokm:
> I'd quite like to make `pyi_builder.test_source()` run the unfrozen code in a subprocess either before the actual test or after if it fails, just so that we can make these upstream breakages easier to triage.

Yeah, that would be useful. There's probably tests where we would need to disable this, though (for example, the ones that can run only as .app bundles on macOS). And the tests that produce result files for the test to read would need to be adjusted, too.

PR #943, opened 2025-09-03
Harden hooks against missing RECORD in distribution's metadata

rokm:
Harden the hooks that iterate over distribution's list of files (i.e.,  `blib2to3` and `torch` hooks). If a distribution's metadata is missing a `RECORD` file (as often seems to be the case with Fedora python packages, for example), the call to `Distribution.files()` returns `None`, and a naive attempt to iterate over the returned value ends up raising a `'NoneType' object is not iterable` error. 

Closes pyinstaller/pyinstaller#9230 (as per https://github.com/pyinstaller/pyinstaller/issues/9230#issuecomment-3246797604).

Update the `clr` hook to use `importlib_metadata` from `_pyinstaller_hooks_contrib.compat` instead of trying to find a suitable version on its own. Modify the guard against `importlib_metadata.files('pythonnet')` returning `None` by using the same pattern as in hooks above; both for the sake of consistency and to remove one level of indentation.

### DIFF
--- a/_pyinstaller_hooks_contrib/stdhooks/hook-clr.py
+++ b/_pyinstaller_hooks_contrib/stdhooks/hook-clr.py
@@ -20,24 +20,19 @@ if is_module_satisfies("pythonnet"):
     # pythonnet requires both clr.pyd and Python.Runtime.dll, but the latter isn't found by PyInstaller.
     import ctypes.util
     from PyInstaller.log import logger
-
-    try:
-        import importlib.metadata as importlib_metadata
-    except ImportError:
-        import importlib_metadata
+    from _pyinstaller_hooks_contrib.compat import importlib_metadata
 
     collected_runtime_files = []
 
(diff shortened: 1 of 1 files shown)


bwoodsend:
I guess we're just writing off the case where the record is missing but we did need to collect some of its files?

rokm:
Yeah. 

The `clr` hook already did that. 

For `torch`, I think we are mostly addressing a hypothetical issue, because the use of `Distribution.files()` happens in Windows specific codepath for collecting MKL DLLs. We have separate codepath for Anaconda there, so this would have to be a misbehaving PyPI wheel.

The Fedora-packaged `black` that's triggering the issue in `blib2to3` does not seem to have the offending `mypy` file in the first place. So even if we were to emit a warning about being unable to discover the `mypy` file due to lack of RECORD, it would be a false positive in this case.

PR #962, opened 2025-10-29
Added hook for nicegui

himbeles:


_pyinstaller_hooks_contrib/stdhooks/hook-nicegui.py
```diff
@@ -0,0 +1,14 @@
...
+# License (version 2.0 or later).
+#
+# The full license is available in LICENSE, distributed with
+# this software.
+#
+# SPDX-License-Identifier: GPL-2.0-or-later
+# ------------------------------------------------------------------
+from PyInstaller.utils.hooks import collect_data_files
+
+datas = collect_data_files('nicegui', include_py_files=False)
```
rokm: `include_py_files` is `False` by default, so it is best to omit it altogether.

PR #970, opened 2025-12-05
hooks: cleanup: port old hooks to PyInstaller.isolated

rokm:
Replace uses of old `exec_statement()` and `eval_statement()` hook utility functions with the `PyInstaller.isolated` framework that has been available since PyInstaller 5.0.

In following hooks, the old `exec/eval_statement` was replaced with equivalent isolated function:
- win32com
- wx.lib.activex
- enchant
- sentry_sdk
- ffpyplayer

In `OpenGL` hook, use standard `collect_submodules` hook utility function to collect submodules of `OpenGL.arrays` instead of trying to reinvent the wheel.

Remove the `gst._gst` hook, which was catering `pygst`, the python-2.7-era GStreamer bindings.


rokm:
Here are the oneshots with affected packages installed (although I don't think we have tests for all of them):
- windows: https://github.com/rokm/pyinstaller-hooks-contrib/actions/runs/19978567262
- macOS: https://github.com/rokm/pyinstaller-hooks-contrib/actions/runs/19978602799
- linux: https://github.com/rokm/pyinstaller-hooks-contrib/actions/runs/19978612396

PR #972, opened 2025-12-12
Add hook for tkinterweb-tkhtml-extras

Andereoo:
Hopefully everything is in order.

This will be the last one, I promise!

### DIFF
--- a/news/972.new.rst
+++ b/news/972.new.rst
@@ -1,1 +1,1 @@
-Add a hook for ``tkinterweb-tkhtml-extras``, which has hidden files.
+Add a hook for ``tkinterweb-tkhtml-extras``.


rokm:
Hmm, actually - do we even need an entry in `requirements-test-libraries.txt`? Does installing `tkinterweb-tkhtml-extras
` in addition to `tkinterweb` (and `tkinterweb-tkhtml`) changes anything w.r.t to [the tests that we already have](https://github.com/pyinstaller/pyinstaller-hooks-contrib/blob/cf899553f81da6c949b725f231469b0a7c9729bc/tests/test_libraries.py#L2908-L2945)? 

Would it be possible to set up a copy of `test_tkinterweb_tkhtml` that uses `tkinterweb-tkhtml-extras`?

rokm:
Ah, I see that if  `tkinterweb-tkhtml-extras` is available but its files are not collected, we get a `FileNotFoundError` when trying to import `tkinterweb_tkhtml` (which tries to list both its files and those from `extras`).

PR #988, opened 2026-02-17
ci: update actions

rokm:
Update used actions to their latest versions:
- `actions/checkout@v4` -> `actions/checkout@v6`
- `actions/setup-python@v5` -> `actions/setup-python@v6`
- `actions/setup-dotnet@v4` -> `actions/setup-dotnet@v5`

Update the `release` workflow to fix compatibility issues with contemporary `setuptools` versions (i.e., unpin `towncrier` from old version that uses `pkg_resources`).

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -17,7 +17,7 @@ jobs:
       - name: Setup Python
         uses: actions/setup-python@v6
         with:
-          python-version: 3.9
+          python-version: 3.x
 
       - name: Log into the gh CLI
         run: echo ${{ secrets.GITHUB_TOKEN }} | gh auth login --with-token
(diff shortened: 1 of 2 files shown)


rokm:
> Hmm, I just noticed that we also have the 22.8.0 pin in the main repo's [`requirements-developer.txt`](https://github.com/pyinstaller/pyinstaller/blob/42c16f919ce9be424fd33ce926a2bbce05c42430/tests/requirements-developer.txt#L23), while the [validation workflow](https://github.com/pyinstaller/pyinstaller/blob/develop/.github/workflows/validate-new-news.yml) does not use a pin. How do you install `towncrier` when making a PyInstaller release?

Ahh.. the release script installs it through [`doc/requirements.txt`](https://github.com/pyinstaller/pyinstaller/blob/develop/doc/requirements.txt#L28), where it is currently pinned to 23.11.0.

Do we have any use for the `requirements-developer.txt` in the main repo? If not, we should be getting rid of it. And if we're happy with 23.11.0 for PyInstaller's changelog, we could also use it here.

PR #1005, opened 2026-04-02
Added hook for `tensorrt`

alysenko-nn:
Added new hook for `tensorrt`, which is loading dynamic libs for the correct initialization of the library


rokm:
If shared libraries from `tensorrt_libs` need to be collected, then instead of hook for `tensorrt`, a hook for `tensorrt_libs` with the following contents should do the trick:

```python
# hook-tensorrt_libs.py
from PyInstaller.utils.hooks import collect_dynamic_libs, PY_DYLIB_PATTERNS

binaries = collect_dynamic_libs(
    'tensorrt_libs',
    search_patterns=PY_DYLIB_PATTERNS + ['*.so.*'],
)
``` 

(by default, `collect_dynamic_libs()` collects only `*.so` files on linux, so we need to add `*.so.*` to `search_patterns` to collect versioned libraries).

Then there's no need to check if `tensorrt_libs` is importable (the hook is called only if the package exists). And there also don't seem to be any need to add `tensorrt_libs` to `hiddenimports`, because it is imported directly in `tensorrt_bindings/__init__.py` (which in turn is imported from `tensorrt/__init__.py`).

_pyinstaller_hooks_contrib/stdhooks/hook-tensorrt_libs.py
```diff
@@ -0,0 +1,21 @@
...
+#
+# SPDX-License-Identifier: GPL-2.0-or-later
+# ------------------------------------------------------------------
+
+import sys
+
+from PyInstaller.utils.hooks import PY_DYLIB_PATTERNS, collect_dynamic_libs
+
+_search_patterns = list(PY_DYLIB_PATTERNS)
+if sys.platform.startswith("linux"):
```
rokm: You can use `PyInstaller.compat.is_linux` for this check.

requirements-test-libraries.txt
```diff
@@ -201,6 +201,8 @@ limits==5.8.0; python_version >= "3.10"
 great-expectations==1.15.1; python_version >= "3.10" and python_version < "3.14"
 # Starting with tensorflow 2.17.0, macOS wheels are provided only for arm64 (x86_64 is deprecated).
 tensorflow==2.21.0; python_version >= "3.10" and python_version < "3.14" and (sys_platform != "darwin" or platform_machine == "arm64")
+# TensorRT wheels are platform-specific and typically require NVIDIA drivers/CUDA at runtime.
+tensorrt==10.16.0.72; sys_platform == "linux" and platform_machine == "x86_64" and python_version >= "3.10" and python_version < "3.14"
```
rokm: Would the test also pass on Windows? (That does seem to be the case in my Windows VM). In that case, the package could be installed if `sys_platform` is either `linux` or `win32`.

_pyinstaller_hooks_contrib/stdhooks/hook-tensorrt_libs.py
alysenko-nn: Thanks @rokm

requirements-test-libraries.txt
alysenko-nn: Make sense, removed sys-platform conditions

bwoodsend: It still needs the platform conditions but to exclude macOS rather than only allow Linux. Something like `and (sys_platform == "linux" or sys_platform == "windows")`

rokm (approved):

rokm:
Thank you!

PR #1008, opened 2026-04-17
Add pyinstaller hook for plum-dispatch

jessamynsmith:


_pyinstaller_hooks_contrib/stdhooks/hook-plum_dispatch.py
```diff
@@ -0,0 +1,18 @@
...
+# this software.
+#
+# SPDX-License-Identifier: GPL-2.0-or-later
+# ------------------------------------------------------------------
+from _pyinstaller_hooks_contrib.utils.mypy import find_mypyc_module_for_dist
+from PyInstaller.utils.hooks import collect_all
+
+# plum-dispatch uses `mypyc`, and includes a top-level module with
+# dynamically-generated name prefix; e.g., `ddc459050edb75a05942__mypyc`.
+datas, binaries, hiddenimports = collect_all('plum')
```
rokm: Is this `collect_all` required at all?

jessamynsmith: Thank you, I hadn't been able to work around it originally and hadn't tried again after finding find_mypyc_module_for_dist. It is now replaced with collect_submodules.

jessamynsmith:
@rokm The version of plum-dispatch I am using requires python 3.10 or greater. Do hooks need to support 3.8 and 3.9 as well?

rokm:
> @rokm The version of plum-dispatch I am using requires python 3.10 or greater. Do hooks need to support 3.8 and 3.9 as well?

In this repository, we test only against the latest version of a package (on the versions of python it supports). So you can change

```
plum-dispatch==2.8.0
```

into 

```
plum-dispatch==2.8.0; python_version >= "3.10"
```

(same as we had elsewhere in the `requirements-test-libraries.txt`.

jessamynsmith:
Thank you, PR is updated.

_pyinstaller_hooks_contrib/stdhooks/hook-plum_dispatch.py
rokm: I'm not sure if `collect_submodules()` is required at all, either - I can get test pass (at least on linux) by just collecting the `mypyc` file. 

Do you get a missing-module error without explicitly collecting all submodules (in your actual use-case, not necessarily the test)?

jessamynsmith: Earlier today I had a version without collect_submodules that was giving an error in my actual use-case (though I was unable to make a test replicating it). However, I have removed it in the most recent commit and my project is still building so I think it's fine without collect_submodules.

bwoodsend (approved):

PR #1011, opened 2026-04-22
Add hook for psycopg[c]

### LINKED ISSUE #1010: Add hook for `psycopg[c]`

`psycopg[c]`

Yes, but it needs several hidden imports.

**Additional context**

This repo already ships [`hook-psycopg_binary.py`](https://github.com/pyinstaller/pyinstaller-hooks-contrib/blob/master/_pyinstaller_hooks_contrib/stdhooks/hook-psycopg_binary.py) for the `psycopg[binary]` wheel variant, but there is no equivalent for `psycopg[c]`. Both variants share the same Cython source and hit the same static-analysis miss.

aeroyorch:
Closes #1010


rokm (approved):

rokm:
Thank you!

PR #1021, opened 2026-05-21
Fix bug setting GDAL_DATA in `osgeo` runtime hook

### LINKED ISSUE #1020: Outdated logic in osgeo runtime hook
**Describe the bug**
`rthooks/pyi_rth_osgeo.py` looks for a specific data file (`gcs.csv`) which is no longer included by the `osgeo/gdal` library.

**Does the error get raised while building or when running?**
There is no error. This hook sets the environment variable `GDAL_DATA`, but in this case, the variable does not get set.

There are runtime warnings from an executable, such as,
```
osgeo              WARNING  [errno 3] Cannot find header.dxf (GDAL_DATA is not defined)
```

**To Reproduce**
On Windows,
```
mamba create ./env
mamba activate ./env
mamba install gdal
ls ./env/Library/share/gdal
```
Observe that `gcs.csv` does not exist among datas that will be collected by `stdhooks/hook-osgeo.py`

**Expected behavior**
`GDAL_DATA` environment variable should be set to `os.path.join(sys._MEIPASS, 'Library', 'share', 'gdal')`

The runtime hook should not check for `gcs.csv` because it no longer exists in that folder:
```python
    gdal_data = os.path.join(sys._MEIPASS, 'Library', 'share', 'gdal')
    # last attempt, check if one of the required file is in the generic folder Library/data
    if not os.path.exists(os.path.join(gdal_data, 'gcs.csv')):
        gdal_data = os.path.join(sys._MEIPASS, 'Library', 'data')
```

 - OS: Windows
 - Python Version: 3.13
 - Version of `pyinstaller-hooks-contrib`: 2026.5
 - Version of PyInstaller: 6.20.0

**Additional context**
I am willing to submit a fix for this, either to look for some other file besides `gcs.csv`, or to simply check if `os.path.join(sys._MEIPASS, 'Library', 'share', 'gdal')` exists without checking for any specific files within.

davemfish:
Fixes bug in `osgeo` runtime hook where the `GDAL_DATA` environment variable was not being set on Windows with a conda environment.

Fixes #1020 

I tested these changes locally by building my project and running its exe with this hook.

I also added an assertion to the existing `osgeo` test in `test_libraries.py`, but `gdal` is missing from `requirements-test-libraries.txt`, so I'm not sure if/when that test will run. GDAL is tough to `pip install` so perhaps that's why it is not in the requirements.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,3 +1,15 @@
+Unreleased Changes
+------------------
+
+Updated hooks
+~~~~~~~~~~~~~
+
+* Fix bug in ``pyi_rth_osgeo.py`` where the "GDAL_DATA" environment variable
+  would not get set on Windows when building from a conda environment.
+  (`#1020
+  <https://github.com/pyinstaller/pyinstaller-hooks-contrib/issues/1020>`_)
+
+
 2026.5 (2026-05-04)
 -------------------
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,15 +1,3 @@
-Unreleased Changes
-------------------
-
-Updated hooks
-~~~~~~~~~~~~~
-
-* Fix bug in ``pyi_rth_osgeo.py`` where the "GDAL_DATA" environment variable
-  would not get set on Windows when building from a conda environment.
-  (`#1020
-  <https://github.com/pyinstaller/pyinstaller-hooks-contrib/issues/1020>`_)
-
-
 2026.5 (2026-05-04)
 -------------------
 
--- a/_pyinstaller_hooks_contrib/rthooks/pyi_rth_osgeo.py
+++ b/_pyinstaller_hooks_contrib/rthooks/pyi_rth_osgeo.py
@@ -27,7 +27,7 @@ if is_win:
             generic_folder = os.path.join(sys._MEIPASS, 'Library', 'data')
             # 'gcs.csv' is included by gdal <= 2.4.4
             # 'gdalicon.png' is included by gdal >= 2.4.4 (and presumeably earlier)
-            if os.path.exists(os.path.join(generic_folder, 'gcs.csv')) | \
+            if os.path.exists(os.path.join(generic_folder, 'gcs.csv')) or \
                     os.path.exists(os.path.join(generic_folder, 'gdalicon.png')):
                 gdal_data = generic_folder
 
--- a/_pyinstaller_hooks_contrib/rthooks/pyi_rth_osgeo.py
+++ b/_pyinstaller_hooks_contrib/rthooks/pyi_rth_osgeo.py
@@ -21,9 +21,13 @@ if is_win:
     if not os.path.exists(gdal_data):
 
         gdal_data = os.path.join(sys._MEIPASS, 'Library', 'share', 'gdal')
-        # last attempt, check if one of the required file is in the generic folder Library/data
-        if not os.path.exists(os.path.join(gdal_data, 'gcs.csv')):
-            gdal_data = os.path.join(sys._MEIPASS, 'Library', 'data')
+        if not os.path.exists(gdal_data):
+            # last attempt, check if one of the required files is in the
+            # generic folder Library/data
+            generic_folder = os.path.join(sys._MEIPASS, 'Library', 'data')
+            if os.path.exists(os.path.join(
+                    generic_folder, 'gdalinfo_output.schema.json')):
+                gdal_data = generic_folder
 
 else:
     gdal_data = os.path.join(sys._MEIPASS, 'share', 'gdal')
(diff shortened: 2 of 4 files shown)


_pyinstaller_hooks_contrib/rthooks/pyi_rth_osgeo.py
```diff
@@ -21,9 +21,13 @@
...
         gdal_data = os.path.join(sys._MEIPASS, 'Library', 'share', 'gdal')
-        # last attempt, check if one of the required file is in the generic folder Library/data
-        if not os.path.exists(os.path.join(gdal_data, 'gcs.csv')):
-            gdal_data = os.path.join(sys._MEIPASS, 'Library', 'data')
+        if not os.path.exists(gdal_data):
+            # last attempt, check if one of the required files is in the
+            # generic folder Library/data
+            generic_folder = os.path.join(sys._MEIPASS, 'Library', 'data')
+            if os.path.exists(os.path.join(
+                    generic_folder, 'gdalinfo_output.schema.json')):
```
rokm: The wrap here is unnecessary, as we have 120 character line limits.

rokm: Was `gdalinfo_output.schema.json` also present in earlier versions where `gcs.csv` existed? If not, we should be checking for either (unless those versions are not supported by python 3.8 and later).

_pyinstaller_hooks_contrib/rthooks/pyi_rth_osgeo.py
```diff
@@ -21,9 +21,13 @@
...
+                gdal_data = generic_folder
```
rokm, line 32: `gdal_data` might end up unassigned in the Windows codepath now.

So it would probably be best to initialize it with `None` at the top of the if-block, and change the `if os.path.exists(gdal_data):` check at the bottom of the hook into `if gdal_data and os.path.exists(gdal_data):`

CHANGELOG.rst
```diff
@@ -1,3 +1,15 @@
+Unreleased Changes
+------------------
+
+Updated hooks
+~~~~~~~~~~~~~
+
+* Fix bug in ``pyi_rth_osgeo.py`` where the "GDAL_DATA" environment variable
+  would not get set on Windows when building from a conda environment.
+  (`#1020
+  <https://github.com/pyinstaller/pyinstaller-hooks-contrib/issues/1020>`_)
```
rokm: CHANGELOG.rst should not be edited directly.  Instead, you should add a news fragment file (`news/1021.update.rst`) that describes your change. `towncrier` will add the reference to PR, so you don't have to explicitly reference issue/PR number in the added text, either.

_pyinstaller_hooks_contrib/rthooks/pyi_rth_osgeo.py
davemfish: This json file was not present alongside `gcs.csv`, so there could have been versions where neither were present. I updated this to check for either `gcs.csv` or `gdalicon.png` because I confirmed that both were present in version 2.4.4, and the png has been there ever since.

I'm not a GDAL maintainer so I don't really know what is the best choice here. We could try to get them to weigh in if necessary.

_pyinstaller_hooks_contrib/rthooks/pyi_rth_osgeo.py
davemfish: `gdal_data` is assigned as `gdal_data = os.path.join(sys._MEIPASS, 'data', 'gdal')` at the top of the `is_win` code block, so I don't think it can ever be `None`.

CHANGELOG.rst
davemfish: Thanks, I guess I didn't read the README closely enough. Reverted this change and added the `news/1021.update.rst` file.

_pyinstaller_hooks_contrib/rthooks/pyi_rth_osgeo.py
rokm: Ah yes, you are correct. Not sure how I manage to see it missing.

_pyinstaller_hooks_contrib/rthooks/pyi_rth_osgeo.py
```diff
@@ -21,9 +21,15 @@
...
-        # last attempt, check if one of the required file is in the generic folder Library/data
-        if not os.path.exists(os.path.join(gdal_data, 'gcs.csv')):
-            gdal_data = os.path.join(sys._MEIPASS, 'Library', 'data')
+        if not os.path.exists(gdal_data):
+            # last attempt, check if one of the required files is in the
+            # generic folder Library/data.
+            generic_folder = os.path.join(sys._MEIPASS, 'Library', 'data')
+            # 'gcs.csv' is included by gdal <= 2.4.4
+            # 'gdalicon.png' is included by gdal >= 2.4.4 (and presumeably earlier)
+            if os.path.exists(os.path.join(generic_folder, 'gcs.csv')) | \
```
rokm: Strictly speaking, this should be a logical `or` rather than bitwise `|`.

davemfish: Thank you! Updated.

rokm (approved):

rokm:
Thank you! 

I've tried running the test, and it fails on `assert(sr.EPSGTreatsAsLatLong())` - presumably because PROJ data is not collected. So we'll probably need to extend the hook to also collect `Library\share\proj` and the run-time hook to set `PROJ_DATA`...

PR #1027, opened 2026-06-12
Fix typo in clone command URL in README

EfazDev:
fixed a quick typo in the README where the command was misspelled with three "o"s.


bwoodsend:
Thanks. Surprised it's lasted that long

PR #1036, opened 2026-07-23
Hook for procrastinate

art321ur:
`procrastinate` needs two things that PyInstaller does not pick up by itself:

  - its SQL data files (`sql/queries.sql`, `sql/schema.sql` and the migrations),
    which it reads via `importlib.resources`
  - its own distribution metadata, which it reads via `importlib.metadata`
    when the package is imported

### DIFF
--- a/requirements-test-libraries.txt
+++ b/requirements-test-libraries.txt
@@ -90,6 +90,10 @@ pinyin==0.4.0
 platformdirs==4.10.1; python_version >= "3.10"
 plotly==6.9.0
 plum-dispatch==2.9.0; python_version >= "3.10"
+procrastinate==3.9.0; python_version >= "3.10"
+# procrastinate depends on psycopg[pool]; the binary extra is required so that psycopg can be
+# imported without a system-provided libpq.
+psycopg[binary]==3.3.4; python_version >= "3.10"
 publicsuffix2==2.20191221
 pycparser==3.0; python_version >= "3.10"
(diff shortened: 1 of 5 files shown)


requirements-test-libraries.txt
```diff
@@ -90,6 +90,10 @@ pinyin==0.4.0
 platformdirs==4.10.1; python_version >= "3.10"
 plotly==6.9.0
 plum-dispatch==2.9.0; python_version >= "3.10"
+procrastinate==3.9.0; python_version >= "3.10"
+# procrastinate depends on psycopg[pool]; the binary extra is required so that psycopg can be
+# imported without a system-provided libpq.
+psycopg[binary]==3.3.4; python_version >= "3.10"
```
rokm: This will not work with our diff-based dependency updates and tests, because the entries for `procrastinate` and `psycopg[binary]` may not end up in the same diff. 

Instead of adding a `psycopg[binary]` dependency here, an install rule like this

``` 
pip show -qq procrastinate && pip install --prefer-binary psycopg[binary]
```

should be added to
https://github.com/pyinstaller/pyinstaller-hooks-contrib/blob/0d1da5a68ddaa1e15195b41b675d600a33432c65/.github/workflows/pr-test.yml#L167-L193

rokm (approved):

rokm:
Thank you!

PR #1044, opened 2026-09-10
hooks: add hook for fastmcp

cz6ace:
- hook
- tests


_pyinstaller_hooks_contrib/stdhooks/hook-fastmcp.py
```diff
@@ -0,0 +1,17 @@
...
+# this software.
+#
+# SPDX-License-Identifier: GPL-2.0-or-later
+# ------------------------------------------------------------------
+
+# Hook for the fastmcp module: https://github.com/PrefectHQ/fastmcp
+
+from PyInstaller.utils.hooks import copy_metadata
+
+datas = copy_metadata('fastmcp')
```
rokm, line 17: Based on [how `fastmcp` attempts to obtain its version information](https://github.com/PrefectHQ/fastmcp/blob/main/fastmcp_slim/fastmcp/__init__.py#L35-L38), the hook should probably try to collect metadata for `fastmcp-slim` and then fall back to collecting metadata for `fastmcp` (or collect both of them if they are available).

news/1044.new.rst
```diff
@@ -0,0 +1 @@
+Add a hook for ``fastmcp``, which has a metadata.
```
rokm, line 1: Most (if not all) dists have metadata. The reason why it needs to be collected here is that the package tries to obtain its version from the said metadata.

(But as far as I'm concerned, an `Add a hook for ``fastmcp``.` would suffice here).

cz6ace: @rokm , good catch. If you agree, I'd collect both fastmcp and fastmcp-slim

rokm:
> @rokm , good catch. If you agree, I'd collect both fastmcp and fastmcp-slim

Yeah, that seems reasonable.

But `copy_metadata()` will end up throwing `PackageNotFoundError` (from either `importlib_metadata` or `importlib.metadata`, depending which is used... or in PyInstaller 5.x even the counterpart from `pkg_resources`) when a dist is not found. 

So assuming that either can be missing, this collection should be made optional, for example:

```python
# See https://github.com/PrefectHQ/fastmcp/blob/v4.0.4/fastmcp_slim/fastmcp/__init__.py#L35-L38
datas = []
for dist_name in ['fastmcp-slim', 'fastmcp']:
    try:
        datas += copy_metadata(dist_name)
    except Exception:  # support both PyInstaller 6.x and 5.x
        pass
```

cz6ace:
@rokm , agreed, fixed, squashed.

rokm:
> @rokm , agreed, fixed, squashed.

Great, thanks!

I've taken a liberty of rephrasing the news fragment a bit, and updated the pinned version to 4.0.4, which seems to have been released yesterday.

rokm (approved):

